PR #214, opened 2013-04-02
Issue #211: Ensure the DateTime is UTC

kojoru:
This commit should allow EnsureUtc mode in SQLServerOrmLiteDialectProvider which will convert DateTime to UTC if it is DateTimeKind.Local before saving. 
The returned DateTimes will get DateTimeKind.Utc so that occasional conversion to UTC will not corrupt it.


bgiromini:
It seems that this feature does not exist in the current iteration of the source code. Is there way it could get incorporated?

bgiromini:
NM, I scoured the source code and see the code was deprecated and replaced with

 `dialect.GetDateTimeConverter().DateStyle = DateTimeKind.Utc; `

PR #430, opened 2015-01-02
Make updating blobs work for Oracle and SQLite.

BruceCowan-AI:
Updating blobs was failing because the data was written converted to a base64
string and read as binary.


mythz:
Cheers, thx Bruce :+1:

PR #431, opened 2015-01-26
AddOracleReservedWordsAndSupportForDateTimeOffset

TroyDycavinu-AI:
-Added more reserved words to Oracle reserved word list
-Updated unit tests for new reserved words
-Added DateTimeOffset (date time with timezone) support for oracle provider
-Added unit tests for querying timestamp values

### DIFF
--- a/src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
+++ b/src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Hosting;
 using System.Data;
 using System.Linq;
 using NUnit.Framework;
@@ -17,20 +20,7 @@ namespace ServiceStack.OrmLite.Tests
             db.CreateTable<ParamTestBo>(true);
             db.CreateTable<ParamRelBo>(true);
             db.CreateTable<ParamByteBo>(true);
-            db.CreateTable<ParamComment>(true);
-            db.CreateTable<ParamOrder>(true);
-            db.CreateTable<ParamLeft>(true);
-            db.CreateTable<ParamUser>(true);
-            db.CreateTable<ParamPassword>(true);
-            db.CreateTable<ParamActive>(true);
-            db.CreateTable<ParamDouble>(true);
-            db.CreateTable<ParamFloat>(true);
-            db.CreateTable<ParamDecimal>(true);
-            db.CreateTable<ParamString>(true);
-            db.CreateTable<ParamDate>(true);
-            db.CreateTable<ParamDateTime>(true);
-            db.CreateTable<ParamType>(true);
-            db.CreateTable<ParamTimestamp>(true);
+            db.CreateTable<ParamLevelAsAlias>(true);
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


TroyDycavinu-AI:
@schmidt4brains please review the Oracle DateTimeOffset changes
@jessemcdowell-AI please review the Oracle reserved word changes

mythz:
Sweet PR thx! 

Will merge once @schmidt4brains and @jessemcdowell-AI gives the all clear.

src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
```diff
@@ -232,14 +247,28 @@ public override string GetQuotedValue(object value, Type fieldType)
                 var dateValue = (DateTime)value;
                 string iso8601Format = "yyyy-MM-dd";
                 string oracleFormat = "YYYY-MM-DD";
-                if (dateValue.ToString("yyyy-MM-dd HH:mm:ss.fff").EndsWith("00:00:00.000") == false)
+                if (dateValue.ToString("yyyy-MM-dd HH:mm:ss.fffffff").EndsWith("00:00:00.0000000") == false)
```
schmidt4brains: I think the test for "is this a start-of-day?" can be more concisely expressed by following [this pattern](https://github.com/ServiceStack/ServiceStack.Text/blob/master/src/ServiceStack.Text/Common/DateTimeSerializer.cs#L376-387) from ServiceStack.Text repo.

Something like: `if (dateValue.TimeOfDay.Ticks != 0)` would avoid the intermediate operations of converting to string and parsing the intermediate result.

I think this should work for both `DateTime` and `DateTimeOffset` values.

src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
```diff
@@ -269,328 +259,922 @@ public void ORA_ParamByteTest()
...
+                var row = new ParamLevelAsAlias { Id = 2, ApprovalLevel = 212218 };
                 db.Insert(row);
 
-                row.Comment = 454;
+                row.ApprovalLevel = 676;
                 db.Update(row);
             }
         }
 
+        [ImportMany(typeof(IParam))]
```
jessemcdowell-AI: It looks like you're adding the first usage of System.ComponentModel.Composition in the codebase. I'm not sure how @mythz will feel about that.

src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
```diff
@@ -269,328 +259,922 @@ public void ORA_ParamByteTest()
...
 
         [Test]
-        public void ORA_ReservedNameLeft_Test()
+        public void ORA_ReservedNames_Test()
         {
+            ResolveReservedNameParameters();
             using (var db = OpenDbConnection())
             {
-                DropAndCreateTables(db);
+                foreach (var param in _reservedNameParams)
```
jessemcdowell-AI: If any of these fail, we won't know which one failed without using a debugger. You could use a try / catch that rethrows with the type of param, or use the TestCaseSource attribute to run the method for each param type.

src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
TroyDycavinu-AI: Yes, I used MEF in the Oracle unit tests to resolve all the parameter classes and cut down on code. In the future if we want to test a new parameter we can just implement another class that inherits from IParam and it will automatically be included in the ORA_ReservedNames_Test.

src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
schmidt4brains: The other pattern that might help the readability of the generated SQL would be to test if the value has any fractional seconds (same example from ServiceStack.Text above).

Something like:
```
var hasFractionalSecs = (timeOfDay.Milliseconds != 0) || ((timeOfDay.Ticks%TimeSpan.TicksPerMillisecond) != 0);

if( hasFractionalSecs )
{
    iso8601Format += ".fffffff";
    oracleFormat += ".FF9";
}
```

I don't think this is as important as the start-of-day test, but it certainly will remove a lot of `.000000000` chatter from the debug logs for any times that are human-entered.

schmidt4brains:
That's it for my comments.

src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
TroyDycavinu-AI: There are 117 param types, so I have opted to add the try/catch

TroyDycavinu-AI:
@schmidt4brains @jessemcdowell-AI I have addressed your suggested changes, please review

schmidt4brains:
Looks good here.

src/ServiceStack.OrmLite.Oracle.Tests/OracleParamTests.cs
TroyDycavinu-AI: @mythz is there any issue with me adding System.ComponentModel.Composition to the codebase?

mythz: @TroyDycavinu-AI Yeah adding more deps makes it harder to port, e.g. future PCL profiles, iOS/Android, or CoreCLR builds. Extra deps can also create issues with peoples existing environments, e.g. devs using tools like SmartAssembly, ILMerge, etc. and using obscure parts of .NET Fx is also more likely to have issues running on Mono, which is why I generally prefer to run custom code where possible.

`System.ComponentModel` should already be referenced, do you know if `System.ComponentModel.Composition/Hosting` is in the same .dll?

TroyDycavinu-AI: @mythz System.ComponentModel.Composition is in a different dll. The new dll was added only to the unit tests.

mythz: @TroyDycavinu-AI Cool, totally fine with adding deps to Unit test projects.

jessemcdowell-AI:
Looks good to me too.

TroyDycavinu-AI:
@mythz please review

mythz:
Yep looks good, thx.

TroyDycavinu-AI:
@mythz Thanks Demis

PR #432, opened 2015-02-02
Allow optional parameter list of reference field names to include in LoadSelect()

jklemmack:
Added additional optional string params to LoadSelect methods to allow for selective reference loading.

Example:

    var dbCustomers = db.LoadSelect<Customer>(q => q.Id == customer.Id, "PrimaryAddress");

This is in response to [this UserVoice request](http://servicestack.uservoice.com/forums/176786-feature-requests/suggestions/6656885-allow-loadselect-to-selectively-load-references)

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/ReadExpressionCommandExtensions.cs
+++ b/src/ServiceStack.OrmLite/Expressions/ReadExpressionCommandExtensions.cs
@@ -126,27 +126,27 @@ namespace ServiceStack.OrmLite
             return dbCmd.Scalar<long>(dbCmd.GetDialectProvider().ToRowCountStatement(sql));
         }
 
-        internal static List<T> LoadSelect<T>(this IDbCommand dbCmd, Func<SqlExpression<T>, SqlExpression<T>> expression)
+        internal static List<T> LoadSelect<T>(this IDbCommand dbCmd, Func<SqlExpression<T>, SqlExpression<T>> expression, string[] include = null)
         {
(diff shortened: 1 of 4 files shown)


mythz:
Hi Johann, this looks like a nice feature, thanks.

Could you also approve the [Contributor License Agreement](https://docs.google.com/forms/d/16Op0fmKaqYtxGL4sg7w_g-cXXyCoWjzppgkuqzOeKyk/viewform) so we can merge this, thx.

jklemmack:
Done.  I also contemplated a version that would take a Type list.  Any interest?

mythz:
Normally it would be a good idea to enable a typed SqlExpression, however this should be obsolete with C# 6.0's new `nameof` operator so I don't think it's worth the added API weight.

Also I've removed `params` from the API which inhibits forward API compatibility for any future features we'd want to add later. The effect of this change can be seen in:
https://github.com/ServiceStack/ServiceStack.OrmLite/commit/92817adf23217a055ac86fa6651669a1c1b6aeee

mythz:
FYI, if you need it before-hand, this change just finished [deploying to MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

Cheers!

jklemmack:
So, using `nameof`, we'd do something like this (?)

    var dbCustomers = db.LoadSelect<Customer>(q => q.Id == customer.Id, 
            new string[] { nameof(customer.PrimaryAddress) } );

Love - it compile-time checking of magic strings :)

mythz:
Yep, it's one of the nicer features :)

PR #433, opened 2015-02-23
Demonstrate either a problem in SaveAll or a misunderstanding on my part...

BruceCowan-AI:
....

I think the added test should pass. It fails.


mythz:
Hi Bruce, should now be resolved with this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/50a2f40193ab68cf03102da882e55fd63ebb4b10

Which is available from v4.0.39+ that's [now on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet) if you need it.

BruceCowan-AI:
Hi Demis,

Thanks! I've bypassed the problem for now, so I can just wait until we upgrade.

PR #435, opened 2015-03-11
Update GetTableName Function

lsowen:
1. Schema name and table name should be separated by "." instead of "_".
2. Each portion of the identifier (eg schema name, table name, column name) should be a maximum of 30 bytes.

Source: http://docs.oracle.com/cd/B28359_01/server.111/b28286/sql_elements008.htm#SQLRF00223


mythz:
great, thx for the PR! :+1:

PR #436, opened 2015-03-11
Updated DropForeignKey<T> to use dialect provider

KevinHoward:
Updated static DropForeignKey<T> method to use dialect provider's syntax from the model definition's GetDropForeignKeyConstraints string result instead of using fixed string. Also consolidated multiple dbConn.GetDialectProvider() calls into a single call.

Was getting an invalid syntax near FOREIGN when running this method against SQL Server because the syntax should be "ALTER TABLE [Table] DROP CONSTRAINT [FK];" yet this method had "ALTER TABLE [Table] DROP FOREIGN KEY [FK];".  (https://msdn.microsoft.com/en-us/library/ms189579.aspx).


mythz:
Cheers, thx Kevin.

PR #437, opened 2015-03-26
Test for bad RowCount SQL statement generated by dialect provider if expression has ORDER BY

UriHendler:
Test and fix for issue ServiceStack/Issues#269


UriHendler:
VS insisted on adding the:
```
<ItemGroup>
    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
</ItemGroup>
```
Wasn't meant to be in the commit.

mythz:
Decided to go with a different impl that just clears the `OrderBy()` in a cloned expression which is less fragile than using string manipulation for this:
https://github.com/ServiceStack/ServiceStack.OrmLite/commit/ebf73b54232fc8d54156698a753c4a32612de91f

Thx for the failing test!

PR #439, opened 2015-04-08
Fix compact guid option in Oracle provider to do searches correctly

BruceCowan-AI:
Before this change the select would ignore the index because it forced Oracle to convert the key value to a string. Ouch.


mythz:
Cheers thx Bruce!

PR #440, opened 2015-04-17
LoadSelect fails with column name conflict on join

Layoric:
Added table names to generated join statement to address [issue 275](https://github.com/ServiceStack/Issues/issues/275)

Could you please have a review @mythz ?


mythz:
Thanks Darren, I'll switch over to use the preferred DialectProvider API and run through the CI tests to check if everything still passes.

Layoric:
@mythz  Ahh ok good to know! :+1:

mythz:
FYI, needed the parent table for self-ref fields, fixed in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/262be4aefbfe265e03eb3018336450b6b23de80b

PR #441, opened 2015-06-05
Changed SqlExpression.VisitUnary ExpressionType.Not => do not generat…

mittork:
…e "NOT(BoolCol = TRUE)" instead generate "BoolCol = FALSE"
- UnitTest adjusted to check for that

---
When doing "NOT(a = TRUE)" in MySql, it is an exclusion, which is not very efficient when using an index. It is better to compare to FALSE.

Of course in my expressions I could write ".Where(e => e.BoolCol == false)" instead of ".Where(e => !e.BoolCol)", but this is annoying :-)

May you take it into consideration.....

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -1169,7 +1169,7 @@ namespace ServiceStack.OrmLite
                         return !((bool)o);
 
                     if (IsFieldName(o))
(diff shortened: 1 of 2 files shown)


mythz:
Could you also [approve the Contributor License Agreement](https://docs.google.com/forms/d/16Op0fmKaqYtxGL4sg7w_g-cXXyCoWjzppgkuqzOeKyk/viewform) so we can evaluate this, thx.

mittork:
Done.

PR #444, opened 2015-06-23
Extended OrmLiteResultsFilter to filter on IDbCommand.

josiahpeters:
OrmLiteResultsFilter now has a SqlCommandFilter action similar to
SqlFilter. This allows you to filter on the actual SqlCommand instead of
just the sql statement itself. This can be useful especially if you are
trying to troubleshoot the values that are being inserted as parameters.

I created a basic result filter similar to the existing CaptureSqlFilter
called CaptureSqlCommandFilter. This attempts to preserve the command data
as a new SqlCommandDetails object with parameter values stored in a
Dictionary along with the Sql string.

I created unit tests for the SqlCommandFilter similar to those for
SqlFilter and added similar checks for existing tests that use SqlFilter.

--

This is a product of me digging into OrmLite to discover why SqlGeography
types weren't being inserted into the database. I've build this
SqlCommandFilter as an alternative to the debugging that I did to track
down my issue, hopefully this will help someone in the future figuring out
exactly what values are being added to the IDbCommand parameters.

### DIFF
--- a/src/ServiceStack.OrmLite/OrmLiteResultsFilter.cs
+++ b/src/ServiceStack.OrmLite/OrmLiteResultsFilter.cs
@@ -68,6 +68,8 @@ namespace ServiceStack.OrmLite
         public Func<IDbCommand, long> LastInsertIdFn { get; set; }
 
         public Action<string> SqlFilter { get; set; }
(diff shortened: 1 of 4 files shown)


mythz:
Excellent, Much better! 

Thanks :+1:

PR #445, opened 2015-06-24
(MSSQL) Add support for Table Valued Functions

mikepugh:
I have several Table Valued Functions in my project. Rather than create & manage POCO objects by hand to deal with the return types, I've updated the T4 template to create them. Similar to views, it is disabled by default and can be switched on via the IncludeFunctions boolean.


mythz:
ok great, thx for putting them behind an opt-in flag.

mikepugh:
No prob!

PR #450, opened 2015-07-16
Added new static method Sql.InExpression to support sub-query SqlExpressions

jklemmack:
Allows much more robust Select filtering.  Sub-expression must return a single column.

Can use like this:

    var customersSubFilter = db.From<Customer>().Select(c => c.Id);
    var orderQuery = db.From<Order>().Where(q => Sql.InExpression(q.CustomerId, customersSubFilter));
    var dbOrders = db.Select(orderQuery);

To generate SQL like this:

    SELECT * FROM Orders WHERE CustomerId IN (SELECT Id From Customers)

### DIFF
--- a/src/ServiceStack.OrmLite.Sqlite/SqliteExpression.cs
+++ b/src/ServiceStack.OrmLite.Sqlite/SqliteExpression.cs
@@ -45,14 +45,13 @@ namespace ServiceStack.OrmLite.Sqlite
             args.RemoveAt(0);
 
             var statement = "";
+            var member = Expression.Convert(m.Arguments[1], typeof(object));
+            var lambda = Expression.Lambda<Func<object>>(member);
+            var getter = lambda.Compile();
 
             switch (m.Method.Name)
             {
                 case "In":
-                    var member = Expression.Convert(m.Arguments[1], typeof(object));
-                    var lambda = Expression.Lambda<Func<object>>(member);
-                    var getter = lambda.Compile();
-
                     var inArgs = Sql.Flatten(getter() as IEnumerable);
 
                     var sIn = new StringBuilder();
@@ -63,6 +62,12 @@ namespace ServiceStack.OrmLite.Sqlite
                             base.DialectProvider.GetQuotedValue(e, e.GetType()));
                     }
                     statement = string.Format("{0} {1} ({2})", quotedColName, m.Method.Name, sIn);
+                    break;
+                case "InExpression":
+                    var sqlExpression = getter() as ISqlExpression;
+                    var subSelect = sqlExpression.ToSelectStatement();
+
+                    statement = string.Format("{0} {1} ({2})", quotedColName, "IN", subSelect);
                     break;
                 case "Desc":
                     statement = string.Format("{0} DESC", quotedColName);
--- a/src/ServiceStack.OrmLite/Expressions/Sql.cs
+++ b/src/ServiceStack.OrmLite/Expressions/Sql.cs
@@ -11,6 +11,11 @@ namespace ServiceStack.OrmLite
             return value != null && Flatten(list).Any(obj => obj.ToString() == value.ToString());
         }
 
+        public static bool InExpression<T, TItem>(T value, SqlExpression<TItem> query)
+        {
+            return value != null && query != null;
+        }
+
         public static List<object> Flatten(IEnumerable list)
         {
             var ret = new List<object>();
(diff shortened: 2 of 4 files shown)


mythz:
Hi Johann, thx for the feature but could you also include some tests showing what it enables and help prevent regressions?

jklemmack:
Whoops - forgot to include.  Added simple Orders -> Customers tests - both affirmative & negative.

mythz:
Great, Thanks :+1: 

Tho I might tweak the API name, it would be good if it can just be `Sql.In()` and differentiate with C# overloading.

jklemmack:
Yeah - wasn't a huge fan of the name.  Total brain freeze on not thinking properly about how overloads work - I was worried about somehow casting the params object array to SqlExpression<T> and gave up.

mythz:
FYI, it's now renamed to `Sql.In()` from this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/2b07dd4cf412fce9d7ea173f7c009271870060e7

jklemmack:
Thanks Demis!  Looks like I also broke other test cases - must have gotten lazy & wanted to go home.  I appreciate you cleaning it up!

Just a thought, but couldn't this approach be used to clean up how you do `LoadSelect` methods, along with executing `Merge()` internally?  Ultimately I (personally) envision some way of building more deeply nested object trees from a single OrmLite call, akin to how Entity Framework does via `Include`.

My issue is that I have a wad of "Status" or "Type" enumeration tables.  So, then if I want to query  some customers along with their orders, the order's object isn't complete without manually executing 1+ more queries for things like "OrderStatus" to get the text description.

Instead, I could do something along the lines of:
 `Db.LoadSelect(Db.From<Customer>(), include: new [] { "*", "Orders.OrderStatus" }`

(not a fan of that wildcard, but its a "for example")

mythz:
I don't want to conceptually change how OrmLite works: as a typed API that's closely modelled after SQL whose results get *intuitively mapped* to the returned POCO. 

If you can think of an intuitive API that fits in this paradigm I can evaluate whether it makes sense to add. We can start with a concrete example of tables we want to support as well what the optimal/intuitive result would be, then I can look at what API makes the most sense to support this use-case.

On a side-note have you thought about using Flag Enums with int values that map to ids in code-tables as a way of avoiding additional code-table lookups?

jklemmack:
I'm actually using [Flags] now, so I"m doing a bit of extra work.  Here's the design pattern:

    [Flags]
    public enum OrderStatusEnum {}

    public class OrderStatus {
        public int Id {get; set;}
        public string OrderStatusName {get; set;}
    }

    public class Order {
        ...
        [References(typeof(OrderStatus))]
        public OrderStatusEnum OrderStatusId {get; set;}
        [Reference]
        public OrderStatus {get; set;}
        ...

During DB generation / migration, the `OrderStatus` table is populated with the enumeration items.  That way, business logic coders have statically typed & defined enumerations, and report writers have all the information they need while maintaining 3rd normal form and a known list of enumeration values.  Its a little bit of extra work up front, but it keeps both camps happy.

But maybe leveraging the `Flags` concept is valid here - an overload / default param `IncludeFlags` that will chain down & load any FLAG-based child.

I agree otherwise about the under riding "simplicity" philosophy of OrmLite, and it's a principle reason I ditched EF.

mythz:
> But maybe leveraging the Flags concept is valid here - an overload / default param IncludeFlags that will chain down & load any FLAG-based child.

Not following what this is trying to do, the `Order.OrderStatusId` is already populated with the `OrderStatusEnum`?

jklemmack:
`Order.OrderStatusId` is populated as an `int`, not a string.  When I load the POCO and spit it back out to the users, I'd like to have a human-friendly text description, instead of the DB ID.

In SQL, we have to join to the `OrderStatus` table to get the text description.  In C#, we'd have to load in the entire contents of `OrderStatus` and `Merge` it.  

Am I doing it horribly wrong?

mythz:
You could have a readonly property that returns the string of the enum and use `[IgnoreDataMember]` to hide integer enum properties from being serialized, e.g:

```csharp
[Flags]
public enum OrderStatus
{
    Pending = 1,
    Cancelled = 2,
    PartiallyShipped = 3 
}

class Order
{
    public int Id { get; set; }

    [IgnoreDataMember]
    public OrderStatus OrderStatusId { get; set; }

    public string OrderStatus {
        get { return OrderStatusId.ToString(); }
    }
}
```

There's also the `OrderStatusId.ToString().SplitCamelCase()` extension method to split the casing into different words.

jklemmack:
In general I don't mind serializing additional data, just don't want to be stuck with ONLY integer data. I like the read-only trick.  With this approach I'd still want to persist the enum its own referenced table for the report-writers, but that's a minor matter.

With the pattern above I'm more than happy to shelve the deep-reference-linking idea.

PR #451, opened 2015-07-16
issue with SqlServerDialectProvider CustomSelect with dots and commas and paging

miiihi:
UseAliasesOrStripTablePrefixes function in SqlServerOrmLiteDialectProvider is not handling well custom selects with commas in it, eg ( SELECT DATEADD(d,1,GETDATE()) AS tommorow, CASE WHEN foo IN (1,2) THEN ...). It breaks the sql. 
Simple solution is to use all fields from subquery (with additional RowNum field last as last). That makes UseAliasesOrStripTablePrefixes obsolete.


mythz:
Great thx for the Fixes + Tests :+1:

PR #453, opened 2015-07-21
to be consistent with base method and what the method name implies, O…

yuinlin:
…racleDialectProvider's

GetParameterValue method should not return quoted values.  move logic for returning either bare or
quoted values into the GetValue method.


yuinlin:
@mythz sorry i should have caught the failed SqlExpressionParamTest.  this pr refactors the oracle dialect provider's GetParamValue to be consistent with the base GetParamValue.

mythz:
No worries, thx Yuin.

PR #454, opened 2015-08-06
fix handling of null variable when parameterized statement is enabled

yuinlin:
added unit test for this case


mythz:
Great, thx :+1:

PR #455, opened 2015-08-06
fix test to use incremented ids instead of hardcoded ids

yuinlin:
fix composite index creation error for oracle provider


mythz:
Cheers :+1:

PR #456, opened 2015-08-19
Added test project demoing the use of Spatials

KevinHoward:
Added new test project demonstrating the use of Spatials (both Geometry and Geography) types with OrmLite.  This functionality depends on the Microsoft.SqlServer.Types library from NuGet, plus references to the unmanaged libraries msvcr100.dll & SqlServerSpatials110.dll that are available from Microsoft's site (SQL Server version specific).

I attempted t0 figure out ways to integrate anything into OrmLite, but no approach was pretty. One thought I had about making this functionality more streamlined with OrmLite, was to wrap the Microsoft Libraries (new project) and provide interfaces that user's could inherit so that they don't have to mess with these over complicated libraries, then possibly provide some SqlExpressions for the type specific functions (i.e. STLength()), but this all seemed outside the scope of just illustrating that you can do it.

https://servicestack.uservoice.com/forums/176786-feature-requests/suggestions/9030526-allow-us-to-use-geography-data-type-in-sql-server

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServerSpatialsTest/Properties/AssemblyInfo.cs
+++ b/src/ServiceStack.OrmLite.SqlServerSpatialsTest/Properties/AssemblyInfo.cs
@@ -1,0 +1,24 @@
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("ServiceStack.OrmLite.SqlServerTests.Spatials")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("ServiceStack")]
+[assembly: AssemblyProduct("ServiceStack.OrmLite.SqlServerTests.Spatials")]
+[assembly: AssemblyCopyright("Copyright © ServiceStack 2015")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+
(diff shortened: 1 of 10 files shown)


mythz:
Wow awesome PR! thx Kevin :) 
You've managed to work out how to use the new Converters whilst the commits weren't dry yet :+1: 

Off the top of my head I'm thinking the best way to package it would be in a new NuGet package since it adds additional deps, e.g:

    PM> Install-Package ServiceStack.OrmLite.SqlServer.Converters

Which could be configured with a 1-liner like: 

```csharp
SqlServerConverters.Configure(SqlServerDialect.Provider);
```

Let me know if you're happy with the above approach? I can then create the necessary NuGet pkg for it.

KevinHoward:
I agree with your approach on the NuGet for Converters.

The custom dialect provider was something I was trying to avoid, but without modifying the base classes, and putting hooks in to register custom converters, it wasn't possible.  

I hope I have provided you with something that can make ServiceStack.OrmLite shine when it comes to extensibility. As you can see, it didn't take a lot of code to do this. The hardest part was dealing with the overly complex Microsoft libraries.

Lastly, to my last point, is there a better way to call the customer type functions in SqlExpression syntax? When I attempted to call the column.STLength() function, it throw an exception stating that it was a potentially hazardous call.

mythz:
Hi Kevin, 

Can you quickly check the new commits to make sure I didn't leave anything out. 

The project was missing some references which didn't appear to be needed so I just removed them from the project:
https://github.com/ServiceStack/ServiceStack.OrmLite/commit/145e38a0d8a34bcda83a9c6f9754463913e92141

I've added a new `InitDbParam()` API to be able to remove the custom dialect provider:
https://github.com/ServiceStack/ServiceStack.OrmLite/commit/c9e1f8b25d60b89f85c01b23116d5bce97422bac

The tests are still passing, but let me know if I've left anything out.

KevinHoward:
Worked beautifully Demis!

The missing references are created by the Microsoft.SqlServer.Types NuGet package which make the assumption that the unmanaged DLLs will live in a specific subdirectory. I pulled some of their code into the TestBase and configured it to load them from the System Directory based on the CPU arch. I'm not sure if there is a way to prevent the NuGet packages from creating this.

I may look at moving the loading of the unmanaged dlls into the Converter's constructors. This should make the setup a bit more inline with everything else..

PR #458, opened 2015-08-20
Added support for the HierarchyId data type to the SqlServer.Converters package

KevinHoward:
Added support for the HierarchyId data type to the *SqlServer.Converters* project, included some support for converting to and from NULL values for all 3 *SqlServer.Converters*, and organized/updated the related testing.

Lastly, I have enabled the NuGet package restore option on the project.

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
+++ b/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
@@ -58,6 +58,7 @@
   <ItemGroup>
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="SqlServerConverters.cs" />
+    <Compile Include="SqlServerHierarchyIdTypeConverter.cs" />
     <Compile Include="SqlServerGeographyTypeConverter.cs" />
     <Compile Include="SqlServerGeometryTypeConverter.cs" />
     <Compile Include="SqlServerTypeConverter.cs" />
--- a/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
+++ b/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
@@ -12,6 +12,8 @@
     <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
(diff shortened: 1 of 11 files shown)


mythz:
Wow impressive work, thx Kevin :+1:

KevinHoward:
Your welcome!  Thought that I should make it more complete and provide a data type that many developers don't know about.

jklemmack:
@KevinHoward Thanks for this!  I'm a fan of the complex data types in SQL Server, and I'm glad to see  first-class support for HierarchyId in ServiceStack.  Any chance you (or @mythz and team) will expose the methods that operate on these types as well?

mythz:
@jklemmack It would require evaluating what the most popular use-cases are to see how they can be intuitively mapped to typed API's. Shipping raw support for it where you can use Custom SQL is a good first step to see how it's being used.

I don't want to rush into polluting API's with any SQLServer-version specific API's which aren't available anywhere else since any API's would break portability. It will take some time to see how they can be scoped to be only visible in the SqlServer provider, maybe it can be done via extension methods on SqlExpression, not sure without prototyping different API's for popular use-cases. I personally avoid using RDBMS specific API's unless they offer a lot of value. 

IMO modelling a tree structure in a flat table with ParentId/ChildId fields and optional expanded Path field (if needed) is pretty intuitive and very portable so it's unlikely I'd ever the use the HierarchyId type myself, so I'd want to see a lot of demand for it first.

csakshaug:
Have you guys consider "Preorder Tree Traversal" ? I have a custom OrmLite implementation for multiple dialect providers running some years now. It is working great and it is pretty easy...

http://www.sitepoint.com/hierarchical-data-database-2/

jklemmack:
Ung - forgot that all the current methods are based on a common SqlExpression, not via provider-specific expressions.  I've found some limited use for HierarchyId, particularly with dynamic multi-level structures, but the point I was aiming at wasn't about HierarchyId specifically.  If you are supporting provider-specific data types (including Geography, which is more interesting to me personally), it'd be nice to see support for the methods.

jklemmack:
@csakshaug I'm familiar with the LFT / RGT structure from an ERP project many moons ago.  My information is a bit dated (about a decade old), but as I recall the downside was you needed to update the whole subtree with any insert or reorder.  In that ERP implementation, we replaced LFT/RGT with a "setstring" implementation.  Its actually very similar to Microsoft's HierarchyId in concept, but uses varchar() strings.  It benefits from very fast inserts and similarly fast lookups.

csakshaug:
@jklemmack It is correct that the inserts is slower, but it was bigger problem before I think. The main thing it is working great on multiple providers without any provider-specific implementation.

KevinHoward:
@mythz I agree with your assessment on implementing the SqlServer Type functions through SqlExpression extension methods. I will take a look - time premitting, but no promises.

csakshaug:
@jklemmack "moons ago" :+1:

PR #459, opened 2015-08-31
Here are some fixes for the problems. Unfortunately, not complete.

BruceCowan-AI:
Hi Demis, here is the promised PR.


mythz:
Great thx Bruce!

PR #464, opened 2015-10-05
Implemented UseParameterizeSqlExpressions in SqlServer

dlj:
Straight rip off from Oracle implementation.
Forced String to be Parameterized as NvarChar(MAX) to reduce number of different plans from the same query.
(This part might be a little iffy, but it greatly reduces the size of the plan cache)


mythz:
I haven't committed this approach/impl yet which is unsupported so I don't want anything of the existing parameterized SqlExpression impl to pollute the existing code-path. I've extracted it out in a separate class which is opt-in with `OrmLiteConfig.UseParameterizeSqlExpressions = true` in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/a3625bef9c78a323b3295be80df8b11a52ae5d95

PR #465, opened 2015-10-30
fix handling of sqlserver datetime2 fractional second precision

yuinlin:
datetime2 can hold up to 7 fractional seconds
add SqlServerDateTime2Converter.ToQuotedString override that includes 7 fractional seconds in format
update unit test to test select on datetime2 value that has 7 fractional seconds


mythz:
Great, thx Yuin :+1:

PR #468, opened 2015-11-15
Added support to generate result type classes for SPs.

rsafier:
Use SQL metadata to build result classes if SPs have output columns. 
I added a global flag IncludeSPReturnTypes (default is false) to control the behavior.


mythz:
great, thx :+1:

PR #470, opened 2015-11-16
Improve SP output export, prevent hidden outputs from becoming properties, and prevent duplicate names.

rsafier:
improvements to SP output export after additional testing on real-life SPs.


mythz:
great, appreciated, thx :+1:

PR #471, opened 2015-11-16
Fix SP & Poco to output properties starting with digits correctly.

rsafier:
More practical fixes against "ugly" DB structure.


mythz:
cheers, let me know when you want me to cut a new build.

rsafier:
I have one more little thing and then I'm done for night.

PR #474, opened 2015-11-17
Further refinements on POCO and SP templates

rsafier:
Some additional improvements as I throw more datasets against it to see how it busts.


mythz:
Great thx, fyi I plan on releasing tomorrow morning (in about 7 hours),.

PR #476, opened 2015-12-05
Removed ToString so it uses StringSerializer

devedse:
I removed the ToString() on an enum so the serialization of it can be
done by the configured StringSerializer.
Because some Expression related stuff changed when building with VS2015
Update 1 this resulted in an issue when I wanted my StringSerializer to
behave differently when serializing enums.

See:
http://stackoverflow.com/questions/34099809/expressions-breaking-code-when-compiled-using-vs2015-update-1


mythz:
ok np, looks like it didn't break any tests, so all good.

devedse:
When do you plan on releasing it to Github :)?

mythz:
Releases are usually every 4-6 weeks so it's unlikely there'll be another one this year. But you can get this change from the [pre-release packages on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

PR #478, opened 2016-01-24
Fix CreateParameter in Oracle provider and add missing DefaultValueTests

BruceCowan-AI:
@mythz Here is a fix for parameterized stuff for the Oracle provider. It cuts the failing tests from 37 to 21, and it makes your new parameterized updates work. BTW, thanks very much for that - we really wanted that. Cheers, Bruce


mythz:
Great, thx Bruce.

PR #479, opened 2016-01-24
Add support in template for explicit AutoQuery typed DTOs to be generated

rsafier:
Default behavior will do not generate unless end-user turns flags on. Only tested again SQL 2012 DBs, but quite an array of tables so have caught most of the edge cases.


mythz:
Great Stuff, thx Richard :+1:

PR #482, opened 2016-01-26
Correct typo in new CreateAutoQueryTypes feature of T4 template

rsafier:
Had incorrect property name in template. Corrected to intended name.


mythz:
Cheers :+1:

PR #485, opened 2016-01-27
Fix time stamp types

BruceCowan-AI:
@mythz Here are a couple more fixes for the Oracle provider. It now only fails on the tests that it has always failed on.
It also supports the Oracle fully-managed provider (but defaults to the old one), except for a problem with initialization which I haven't figured out - when it attempts to load the managed provider a callback into OrmLite occurs for some reason which I don't understand and things go very sideways from there because it starts executing code which should not be run at that point in time, If you comment out the oracle.manageddataaccess section in machine.config then it all works. Cheers, Bruce


mythz:
Awesome thx Bruce, 

Out of curiosity what are the steps to use the Managed Provider, is it just a matter of installing the NuGet package and passing "Oracle.ManagedDataAccess.Client" in the constructor? do they need to do anything else?

PR #487, opened 2016-01-28
Set UseThreadStatic=true in Oracle provider. OK because the Oracle cl…

BruceCowan-AI:
…ients

do not support async. This gets rid of configuration error.

@mythz Here is the simple fix to make the CaptureSqlFilter tests work on Oracle managed client.


mythz:
cheers :+1:

PR #490, opened 2016-02-13
add functionality to allow enums to be stored as integer values

JohnLeavey:
Add functionality to allow enums to be stored as integer values without having to declare them as "flags"

1. Create new attribute type EnumAsAttrubute. This works just as a property label - the class has no functionality.

2. In OrmLiteConfigExtensions.cs - GetModelDefinition(), Check for this new attribute, and if included, set treatAsType to the underlying type.

3. In OrmLiteDialectProviderBase.cs - GetConverterBestMatch(), use fieldDef.ColumnType rather than fieldDef.FieldType. ColumnType will return FieldType anyway if TreatAsType is null.

### DIFF
--- a/src/ServiceStack.OrmLite/OrmLiteConfigExtensions.cs
+++ b/src/ServiceStack.OrmLite/OrmLiteConfigExtensions.cs
@@ -113,8 +113,10 @@ namespace ServiceStack.OrmLite
                     ? Nullable.GetUnderlyingType(propertyInfo.PropertyType)
                     : propertyInfo.PropertyType;
 
+                var isEnumAsInt = propertyInfo.HasAttribute<EnumAsIntAttribute>();
+
                 Type treatAsType = null;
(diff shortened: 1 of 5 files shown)


mythz:
Hi John, thx for the PR. 

Can I get you to rename `OrmLiteEnumAsIntAttribute.cs` to match the class name, e.g `EnumAsIntAttribute.cs`, otherwise looks good.

mythz:
actually don't worry about it, I'll move it to ServiceStack.Interfaces with the other metadata attributes.

mythz:
I'm reverting this entire feature as it can't work in SQL Expressions since the type needs to be determined from the Type/Value.

I spent some time creating a custom EnumAsIntConverter but ultimately it can't be used in SQL Expressions since we don't have the Field Definition. If there were tests this could've been worked out a lot sooner.

PR #491, opened 2016-02-14
add functionality to allow enums to be stored as integer values #2

JohnLeavey:
Resubmission of pull request #490

Add functionality to allow enums to be stored as integer values without having to declare them as "flags"

Create new attribute type EnumAsAttrubute. This works just as a property label - the class has no functionality.

In OrmLiteConfigExtensions.cs - GetModelDefinition(), Check for this new attribute, and if included, set treatAsType to the underlying type.

In OrmLiteDialectProviderBase.cs - GetConverterBestMatch(), use fieldDef.ColumnType rather than fieldDef.FieldType. ColumnType will return FieldType anyway if TreatAsType is null.

SqlExpression.cs
In GetExpression(), check for EnumAsInt attribute and set new EnumAsInt property of EnumMemberAccess as appropriate. Use this new property in VisitBinary().

Ensure that all test in EnumTests pass.

Copy appropriate tests into EnumAsIntTests, using data classes with EnumAsInt attribute and ensure that all pass.

### DIFF
--- a/src/ServiceStack.OrmLite/EnumAsIntAttribute.cs
+++ b/src/ServiceStack.OrmLite/EnumAsIntAttribute.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
-namespace ServiceStack.OrmLite
+namespace ServiceStack.Interfaces
 {
     public class EnumAsIntAttribute: Attribute
     {
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Collections.ObjectModel;
 using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
+using ServiceStack.Interfaces;
 
 namespace ServiceStack.OrmLite
 {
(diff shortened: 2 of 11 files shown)


mythz:
OrmLite wasn't designed to support different behavior at a property level as seen by the ugly hacks littered throughout different code paths in the code-base, breaking encapsulation outside of the Type Converters. Property level behavior is basically incompatible with OrmLite where trying to add more behavior like this will quickly turn into an unmaintainable mess, devaluing the encapsulation, extensibility and customizability of Type Providers.

I also don't believe in the use-case for this feature, the only time I'd imagine it would be useful is if you wanted the same Enum stored as an int in some columns and as a string in others, basically making them incompatible at the database level - basically causing issues for both library and its consumers so I wont be accepting any property-level behaviors like this.

If for some reason you need adhoc behavior that's different for a specific table you can [use a custom JsScope](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/34c0ad4a8debbe7b58ffe1f29d997c2e311e58b2/tests/ServiceStack.OrmLite.Tests/EnumTests.cs#L328), e.g:

```csharp
using (JsConfig.With(treatEnumAsInteger: true))
{
    db.Insert(new TypeWithTreatEnumAsInt { Id = 1, EnumValue = SomeEnumTreatAsInt.Value1 });
}
```

The other alternative being having 2 different enum definitions, one with the attribute and one without.

But I have added support [[EnumAsInt] attribute](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/34c0ad4a8debbe7b58ffe1f29d997c2e311e58b2) at the Enum level which works similar to [Flags] except it only applies to OrmLite which supports a useful use-case where you want the enum stored as an Int in the DB but serialized as a string at the Service Contract.

You can start using `[EnumAsInt]` on enums from v4.0.53 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet)

PR #492, opened 2016-02-18
Fixed handling of Sql.In(empty array) and Sql.In(null)

shift-evgeny:
Sql.In(empty array) and Sql.In(null) are now evaluated as false (i.e. match no rows) rather than failing with an unhelpful error like "System.Data.SqlClient.SqlException : Incorrect syntax near ')'." or NullReferenceException, respectively.


mythz:
Brilliant, thx :+1:

PR #493, opened 2016-02-18
Demonstrate stack overflow in query.

BruceCowan-AI:
@mythz The added test, Can_query_column_as_nullable demonstrates a stack overflow in a query. I suppose you can argue that it should be illegal to call a column nullable when it isn't nullable in the database, but this code works with 4.0.48 and earlier.


mythz:
Thx for the failing tes Brucet! Should now be fixed from this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/a04f47dc60b903e46162b0d3eec6713a073d4342

PR #494, opened 2016-02-19
Fix the Oracle string length default in the string converter

BruceCowan-AI:
@mythz This fixes the mysterious failures I was getting. Something changed so that the length is copied into the parameter as the max size, so it needs to be correct now.
Shortly I'll be able to get back to the memory leak issue.


mythz:
ok great, thx Bruce.

PR #496, opened 2016-02-25
Fix Oracle string converter and one test

BruceCowan-AI:
@mythz Yes, you fixed the problem. Since you did I don't have to debug tomorrow, so here is the fix for my problem - I hadn't tested UseUnicode true before. And I fixed one more test.


mythz:
cheers, thx

PR #497, opened 2016-03-17
MySQL issue with nullable enum + int casting prob

Layoric:
Hey @mythz,

Related to [issue posted in the forums](https://forums.servicestack.net/t/enumasint-field-not-loading-when-nullable-4-0-54/2170), I found a simple solution to show way forward/issue with nullable enum properties casting from integer.

Better solution would probably be done in the Express building in `PropertyInvoker.cs` 

https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/ServiceStack.OrmLite/PropertyInvoker.cs#L37

Expression could build a predicate checking for Enum + Nullable type, eg 

```
Nullable.GetUnderlyingType(propertyInfo.PropertyType) != null
&& Nullable.GetUnderlyingType(propertyInfo.PropertyType).IsEnum)
``` 

Not familiar with Expression building, so was unsure how to build an expression to call equivalent of:

```
var enumType = Nullable.GetUnderlyingType(fieldDef.PropertyInfo.PropertyType);
var enumValue = Enum.ToObject(enumType, value);
``` 

Only tested with MySQL, but might be a problem with other DBs as well. Any ideas?

### DIFF
--- a/src/ServiceStack.OrmLite.MySql.Tests/EnumTests.cs
+++ b/src/ServiceStack.OrmLite.MySql.Tests/EnumTests.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using NUnit.Framework;
+using ServiceStack.DataAnnotations;
 
 namespace ServiceStack.OrmLite.MySql.Tests
 {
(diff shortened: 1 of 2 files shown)


mythz:
Great thx for the tests, I'll probably be tweaking the impl after I debug it.

mythz:
Hey, the conversion just needed to be moved up into the EnumConverter, just done in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/eb3eb8005ac3672efb3b178cdd57df750a99126f

PR #498, opened 2016-03-23
StackOverflowException when FromDbValue() is called with a value type that has no converter registered

shift-evgeny:
Here is a unit test that reproduces the problem. I'm not sure what it _should_ do, but definitely not _that_. :) In v4.0.44 it returned null, which isn't great either, as it masks the problem (lack of a converter). A helpful exception would probably be good here, but I don't know the best place to throw it.


mythz:
thx for the test :+1:  should now be resolved in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/1baf2123624339f4f0630b382a5e6ec9cd075d68

Basically if types don't match let the serializer try coerce it.

This change is available from v4.0.55 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

PR #499, opened 2016-03-29
Implemented an UpdateIncrement API

madaleno:
Implement an UpdateIncrement API similar to UpdateOnly, but that generates SQL to add the value to the existing database field instead of doing attribution, when the field is a numeric field. Other field types still do attribution and we can mix field types.

db.UpdateIncrement(new Person { Age = 5 }, db.From<Person>().Update(p =>p.Age).Where(x => x.FirstName == "Jimi"));
will generate this SQL:
UPDATE "Person" SET "Age" = "Age" + 5 WHERE ("FirstName" = 'Jimi')

This is usefull to increment counter, totals, etc. without having to load the object from the BD first and avoinding concurrency conflicts.

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/WriteExpressionCommandExtensions.cs
+++ b/src/ServiceStack.OrmLite/Expressions/WriteExpressionCommandExtensions.cs
@@ -51,6 +51,38 @@ namespace ServiceStack.OrmLite
             return dbCmd.UpdateOnly(obj, q);
         }
 
+        public static int UpdateIncrement<T>(this IDbCommand dbCmd, T model, SqlExpression<T> fields)
+        {
+            UpdateIncrementSql(dbCmd, model, fields);
+            return dbCmd.ExecNonQuery();
+        }
+
+        internal static void UpdateIncrementSql<T>(this IDbCommand dbCmd, T model, SqlExpression<T> fields)
+        {
+            if (OrmLiteConfig.UpdateFilter != null)
+                OrmLiteConfig.UpdateFilter(dbCmd, model);
+
+            fields.CopyParamsTo(dbCmd);
+
+            dbCmd.GetDialectProvider().PrepareUpdateRowIncrementStatement(dbCmd, model, fields.UpdateFields);
+
+            if (!fields.WhereExpression.IsNullOrEmpty())
+                dbCmd.CommandText += " " + fields.WhereExpression;
+        }
+
+        public static int UpdateIncrement<T, TKey>(this IDbCommand dbCmd, T obj,
+            Expression<Func<T, TKey>> fields = null,
+            Expression<Func<T, bool>> where = null)
+        {
+            if (fields == null)
+                throw new ArgumentNullException("fields");
+
+            var q = dbCmd.GetDialectProvider().SqlExpression<T>();
+            q.Update(fields);
+            q.Where(where);
+            return dbCmd.UpdateIncrement(obj, q);
+        }
+
         public static int UpdateNonDefaults<T>(this IDbCommand dbCmd, T item, Expression<Func<T, bool>> obj)
         {
             if (OrmLiteConfig.UpdateFilter != null)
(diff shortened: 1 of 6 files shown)


mythz:
Looks like a useful addition, thx :+1: 

Tried to think of a better API to signal it only increments numeric values but couldn't come up with a better one that was terse enough to still be useful, so happy to stick with 'UpdateIncrement`, will just need to document its behavior so it's behavior is clear.

madaleno:
Had a fight to myself when naming this.

mythz:
@madaleno Hi Luis, I've decided to rename this API to `UpdateAdd` from this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/da84b37e0c6542720f9a52e20857cec095906f2f

It's more concise (always better) and follows what others are using for this operation, e.g. DynamoDB uses `ADD` in their [UpdateItem API](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) to add to the existing value, instead of replacing it.

This change is available from the latest v4.0.55 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

madaleno:
Great! Shorter is better.

Luis Madaleno
Em 13/04/2016 5:43 PM, "Demis Bellot" <notifications@github.com> escreveu:

mythz:
Hi Luis,

Wanted to let you know that I believe I've found a nicer API for Update API's where we can just use a lambda to specify both which fields to update and their values, so your original example of:

```csharp
db.UpdateAdd(new Person { Age = 5 }, 
    db.From<Person>().Update(p =>p.Age).Where(x => x.FirstName == "Jimi"));
```

Can now be written as:

```csharp
db.UpdateAdd(() => new Person { Age = 5 }, where: x => x.FirstName == "Jimi");
```

So you no longer have to repeat which fields needs updating twice.

I've also removed the other `UpdateAdd` overloads as I think this is much nicer API and I don't know of a use-case that requires the older APIs and I'd prefer not to drag the existing legacy overloads with this new API. So since this API was just released in the last release I think it's better to introduce a breaking change to the new API now than carry the older overloads forever.

BTW if you have a query that needs the old overloads please give me an example and I'll add a new API that can support it.

The new UpdateAdd API was [added in this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/0c35010ac3e6071808da8cf5c2c9e0eab3dfb4da) that's available from v4.0.57 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

madaleno:
Hi Demis,

Great! That's definitely better.

I'm ok with the breaking change. I don't have anything in production with
this code yet.

Thanks.

Demis Bellot <notifications@github.com> escreveu no dia segunda, 23/05/2016
às 17:59:

> Hi Luis,
>
> Wanted to let you know that I believe I've found a nicer API for Update
> API's where we can just use a lambda to specify both which fields to update
> and their values, so your original example of:
>
> db.UpdateAdd(new Person { Age = 5 },
>     db.From<Person>().Update(p =>p.Age).Where(x => x.FirstName == "Jimi"));
>
> Can now be written as:
>
> db.UpdateAdd(() => new Person { Age = 5 }, where: x => x.FirstName == "Jimi");
>
> So you no longer have to repeat which fields needs updating twice.
>
> I've also removed the other UpdateAdd overloads as I think this is much
> nicer API and I don't know of a use-case that requires the older APIs and
> I'd prefer not to drag the existing legacy overloads with this new API. So
> since this API was just released in the last release I think it's better to
> introduce a breaking change to the new API now than carry the older
> overloads forever.
>
> BTW if you have a query that needs the old overloads please give me an
> example and I'll add a new API that can support it.
>
> The new UpdateAdd API was added in this commit
> <https://github.com/ServiceStack/ServiceStack.OrmLite/commit/0c35010ac3e6071808da8cf5c2c9e0eab3dfb4da>
> that's available from v4.0.57 that's now available on MyGet
> <https://github.com/ServiceStack/ServiceStack/wiki/MyGet>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ServiceStack/ServiceStack.OrmLite/pull/499#issuecomment-221031272>
>
-- 
Cumprimentos,

Luis Madaleno
Dir. Técnico @ magnisoft

PR #501, opened 2016-04-27
Enhance T4 templates for SqlServer to include more foreign key support

jordan-ware:
Hit a problem during development where we'd have an item described below:

```
Item {
   OwnerId: 1,
   Owner2Id: 2
}
```
Added the following properties to the partial class to load references:

```
[Reference]
public Owner Owner { get; set; }
[Reference]
public Owner Owner2 { get; set; }
```
However, on doing a Db.LoadSingleById(), we'd end up with this:

```
Item {
   OwnerId: 1,
   Owner: { Id: 1, Name: "Owner" },
   Owner2Id: 2,
   Owner2: { Id: 1, Name: "Owner" }
}
```
The first Owner entity was being loaded into the second.

Adding the [References(typeof(Owner))] attribute to the foreign key field in the POCO generated by the T4 templates resolved the odd behaviour, and brought the POCO's more in line with the code demonstrated on the Ormlite documentation page.

The result is that in the POCO, this:

```
public int? OwnerId { get; set; }
public int? Owner2Id { get; set; }
```
Becomes this:

```
[References(typeof(Owner))]
public int? OwnerId { get; set; }
[References(typeof(Owner))]
public int? Owner2Id { get; set; }
```

I'm not sure if the underlying issue needs to be addressed. It does not affect me any more. It could still be a nuisance for people using the T4 templates for database engines other than SqlServer or SQLite.

### DIFF
--- a/src/T4/OrmLite.Core.ttinclude
+++ b/src/T4/OrmLite.Core.ttinclude
@@ -829,12 +829,12 @@ class SqlServerSchemaReader : SchemaReader
 
 	public override Tables ReadSchema(DbConnection connection, DbProviderFactory factory)
 	{
-		var result=new Tables();
+		var result = new Tables();
 
 		_connection=connection;
 		_factory=factory;
 
-		var cmd=_factory.CreateCommand();
+		var cmd = _factory.CreateCommand();
 		cmd.Connection=connection;
 		cmd.CommandText=TABLE_SQL;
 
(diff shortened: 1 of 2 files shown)


mythz:
Awesome, thx for the PR - yeah you'll need to do this for [multiple self references](https://github.com/ServiceStack/ServiceStack.OrmLite#multiple-self-references).

> Also happy to see a someone from my home suburb of Nightcliff :)

jordan-ware:
Hahaha yep. We've even met - when you came around to Radical Systems. Can't remember how long ago now though.

No worries about the PR. Enjoying the API!

mythz:
Ahh nice :+1:  yeah Adam gave me a heads up someone from radical was going to submit an issue - but PR's are even more welcome :)

PR #502, opened 2016-05-02
Update T4 template references feature to be optional by flag, and correctly work on Singularized table renames.

rsafier:
Correct T4 Template for `bool MakeSingular = true;` condition, also make it optional to emit Reference attributes (same as all the other options)


mythz:
Cheers 👍

PR #505, opened 2016-05-26
Implement new public IHasDbTransaction interface from ServiceStack.Common

shift-evgeny:
This allows the real transaction to easily be extracted from an OrmLiteTransaction. The existing internal interface was renamed to ICanSetDbTransaction.

These changes will need the new version of ServiceStack.Common.dll to build (not included).


mythz:
I'm going with a slightly the less disruptive alternative to use a less verbose interface name and use the new IHasDbTransaction interface when only readonly is needed in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/fb573afbfe2224df76ded60db6a9d2389b5452d2

shift-evgeny:
Cool, that's even better. Thanks!

PR #506, opened 2016-06-02
Unit test fixes

shift-evgeny:
Some fixes to unit tests that failed for me, especially due a case-sensitive server collation for SQL Server.

PostgreSQL tests still fail due to the mismatch in Npgsql version (2.2.7 vs 3.0.5), which can be fixed by downgrading the test projects to Npgsql 2.2.7, but I'm not including that in this PR.

### DIFF
--- a/src/ServiceStack.OrmLite.PostgreSQL.Tests/CreatePostgreSQLTablesTests.cs
+++ b/src/ServiceStack.OrmLite.PostgreSQL.Tests/CreatePostgreSQLTablesTests.cs
@@ -110,9 +110,9 @@ namespace ServiceStack.OrmLite.PostgreSQL.Tests
 BEGIN
 
     IF NOT EXISTS(
-        SELECT schema_name
-          FROM information_schema.schemata
-          WHERE schema_name = '{0}'
+        SELECT 1
+          FROM INFORMATION_SCHEMA.SCHEMATA
+          WHERE SCHEMA_NAME = '{0}'
       )
     THEN
       EXECUTE 'CREATE SCHEMA ""{0}""';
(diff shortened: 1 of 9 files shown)


KevinHoward:
@shift-evgeny can you also update the SqlHierarchyId type converter column definition in SqlServerHierarchyIdTypeConverter.cs.

KevinHoward:
@shift-evgeny, if the upper case column definitions for the Geography/Geometry types are an issue for a SQL Server with collation set to case-sensitive, wouldn't you have the same issue with all of the other data type column definitions? I ask because all of the other converters have upper case column definitions, yet your commit only updates these two (ex. [FloatConverter](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/ServiceStack.OrmLite/Converters/FloatConverters.cs) & [DateTimeOffset](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/ServiceStack.OrmLite/Converters/DateTimeOffsetConverter.cs)). 

_Sorry, I don't have any databases with collation set to a case-sensitive, I always felt that it's a bad practice, but I understand that sometimes we have to play with the cards that we're dealt._

mythz:
yeah I'm also curious about other types - but I'm more surprised data types are case-sensitive.

shift-evgeny:
I don't think types names are case-sensitive in general. `SELECT CAST('LINESTRING(-1 2, -3 4)' AS GeomeTRY)` works fine on the same SQL Server instance. The geometry/geography thing only fails with the "::" operator - so I guess there is something special about that.

shift-evgeny:
Wait, it's not only with "::". In the OrmLite unit tests it seems that SqlParameter.UdtTypeName is case-sensitive, but creating the table with column `"Shape" GEOmetry NULL` works fine.

PR #508, opened 2016-06-03
Better nullable support in SqlExpression

shift-evgeny:
Several fixes to SqlExpression:

*) Handle "array/list.Contains(x.Value)" where x is a Nullable<T>: equivalent to Contains(x)
*) Handle "x.HasValue" where x is a Nullable<T>: equivalent to "x != null"
*) Handle "null != x": equivalent to "x != null"
*) Handle "new[] { 10, 30 }.Contains" correctly where the array is created inside the lambda body


mythz:
Wow Awesome PR! appreciated, thanks 👍

PR #509, opened 2016-06-08
Handle expressions that evaluate to true/false/nulls in Where() expression

shift-evgeny:
The following now work:

* "Where(x => true)" or "false" or an expression that is evaluated to true or false at runtime (by CachedExpressionCompiler.Evaluate). Of course, this is pointless on its own, but can be useful when building a more complex expression with And(), Or(), etc.
* "Where(x => x.NullableColumn == null)" or an expression that is evaluated to null at runtime
* "Where(x => x.NullableColumn.HasValue = true)" or "false" or an expression evaluated to true/false at runtime
* ... same with the boolean/null expression on the left


mythz:
Sweet, much appreciated :+1:

PR #511, opened 2016-06-09
Add COALESCE binary expression tests

bogdangrigg:
Shows a current issue: when COALESCE is on the right side of a comparison, it's serialized **into** the query param as a string. So the test Can_select_equals_coalesce_on_the_right_expression() will always fail.

From what I see SqlExpression.VisitFilter runs ConvertToPlaceholderAndParameter(ref right). Which adds a param "@1" with the value "{COALESCE("NullableCol",@0)}", then sets right to "@1".

Not sure how to fix it reliably & elegantly... especially to treat nested expressions.


mythz:
ok thx, but we may not fix complicated expressions - there's only so much we can handle converting LINQ expressions into SQL.

mythz:
ok should be fixed with this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/a5cd168f7ab7055857c61a09b858642d09cf69b0

bogdangrigg:
Many thanks Demis!

PR #513, opened 2016-06-23
Made [EnumAsInt] attribute work when applied to the property as well as when applied to the enum type

shift-evgeny:
This is particularly useful when you have no control over the enum type (e.g. System.DayOfWeek)

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -1302,7 +1302,7 @@ namespace ServiceStack.OrmLite
                     var rightPartialSql = right as PartialSqlString;
                     if (rightPartialSql == null)
                     {
-                        right = GetValue(right, leftEnum.EnumType);
+                        right = GetValue(right, leftEnum.ColumnType);
                     }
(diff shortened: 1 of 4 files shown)


mythz:
OrmLite doesn't support Property Type Converters and I don't want to start adding pseudo support for them now because they can't be supported cleanly everywhere it's used and as a result there are going to be bugs filed we can't fix. 

You can still "Add Attributes" to types you don't control with the dynamic attribute API, e.g:

```csharp
typeof(DayOfWeek)
  .AddAttributes(new EnumAsIntAttribute());
```

This needs to be done on Startup before it's used.

shift-evgeny:
Ah, OK. The reason I thought this was supposed to work is that [EnumAsInt] could be applied to properties (`[AttributeUsage(AttributeTargets.Enum | AttributeTargets.Property)]`). Maybe it's worth removing `AttributeTargets.Property` to make it less confusing to users?

Also, is there a way I could contact you to ask about potential changes, so I don't waste time on coding PRs that won't get accepted? The forum is for subscribers only and GitHub issues are disabled in this project.

mythz:
Yeah what's your email?

PR #514, opened 2016-06-24
Select entire table in SqlExpression.Select() using anonymous types

shift-evgeny:
Added support for selecting all of a table's columns in SqlExpression via `.Select<TableA, TableB>((a, b) => new { a })`. If a property name is specified in the anonymous type it's used as a prefix for every column, e.g. `.Select<TableA, TableB>((a, b) => new { ThisIsA = a })` would prefix every column of TableA with "ThisIsA". Property names can also now be used as aliases for individual columns, e.g. `.Select<TableA, TableB>((a, b) => new { a, BName = b.Name })` would result in "SELECT ..., TableB.Name as BName".

I had to add a few extra classes (SelectList, etc.) and make a breaking change the return type of IOrmLiteDialectProvider.GetRowVersionColumnName() to be able to manipulate the column aliases without major hacks. Perhaps they can be useful elsewhere as well. This is technically a breaking change, but hopefully a very minor one - I don't think too many people would be using GetRowVersionColumnName() directly.

### DIFF
--- a/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
@@ -657,10 +657,10 @@ namespace ServiceStack.OrmLite.Firebird
                 .Replace("%", @"^%");
         }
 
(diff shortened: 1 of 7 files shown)


mythz:
Thanks Evgeny 👍

PR #520, opened 2016-07-10
failing test for dynamic api

angelcolmenares:
Porperty name in created dynamic is not the same as Property name passed to select method when using lambdas


mythz:
That's a limitation of using dynamic - it needs to match the returned properties exactly.

PR #522, opened 2016-07-27
Added table hints to Sql Server join expressions

bryancrosby:
It looked like extension methods were the easiest way to implement this. Table hints vary by provider (if any), and putting them on to the base SqlExpression<T> seemed wrong. Open to any suggestions!

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer/ServiceStack.OrmLite.SqlServer.csproj
+++ b/src/ServiceStack.OrmLite.SqlServer/ServiceStack.OrmLite.SqlServer.csproj
@@ -103,7 +103,9 @@
     <Compile Include="SqlServerDialect.cs" />
     <Compile Include="SqlServerExpression.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="SqlServerExpressionExtensions.cs" />
     <Compile Include="SqlServerOrmLiteDialectProvider.cs" />
+    <Compile Include="SqlServerTableHint.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\ServiceStack.OrmLite\ServiceStack.OrmLite.csproj">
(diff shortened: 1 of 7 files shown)


mythz:
Thanks for the PR, I'm likely going to look at making this more generic and useful to all providers by adding an optional string format which all providers can use to customize the join expression. I'll notify you here when I'm done.

bryancrosby:
No problem. I was considering a customization route like you suggest as well, so I think that would be a welcome change that all providers could use. Thanks!

mythz:
FYI, I've made the changes in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/0d8f42994eda40c7b579b4077a75ccc4c5d2b8a5

Your call-site code is near identical except it's just called `Join` instead of `JoinWithHint`, e.g:

```csharp
var join = db.From<Car>()
    .Join<Car, CarType>((l, r) => l.CarId == r.CarId, SqlServerTableHint.ReadUncommitted);
```

Otherwise should work as expected.

This change is available from v4.0.61 that's now [available on MyGet](https://github.com/ServiceStack/ServiceStack/wiki/MyGet).

Thanks for the PR!

mythz:
FYI I've decided to change this to a typed delegate which offers more control and can support more use-cases in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/1a8e895aeceb78d786fd5105a52fe82979f7db33

Your call-site code remains unchanged:

```csharp
var join = db.From<Car>()
    .Join<Car, CarType>((l, r) => l.CarId == r.CarId, SqlServerTableHint.ReadUncommitted);
```

This change is available from v4.0.61 on MyGet, you'll need to [clear your MyGet Cache](https://github.com/ServiceStack/ServiceStack/wiki/MyGet#redownloading-myget-packages) if you already have v4.0.61 installed.

bryancrosby:
Looks great. Thanks for the quick response!

bryancrosby:
FYI I think it might also be useful to have it on the From<T> method as well. Any thoughts on that?

mythz:
It's not a JOIN alias so wont have the same signature, also we don't have the same control over the FROM clause as we do with table joins and it will complicate the intelli-sense for `db.From<T>()` for everyone whilst only benefiting very few, so going to avoid it for now.

PR #523, opened 2016-07-29
Added support "ToString" method in SqlExpression.

OlegNadymov:
I added support"ToString" method in SqlExpression. 
Now it's possible to use expressions like that:
```C#
var q = Db.From<TestType>().Where(x => x.NullableIntCol.ToString().EndsWith("0"));
```


mythz:
Looks good, Cheers :+1:

PR #524, opened 2016-07-29
Added support Concat string in SqlExpression

OlegNadymov:
I added support "String.Concat" method in SqlExpression.
Now it's possible to use expressions like that:
```C#
var q = Db.From<TestType>().
Where(xx => String.Concat(x.TextCol, x.NullableIntCol.ToString()) =="anyString");
```
Also I overrode processing for operator "+" for string arguments. Now below expression works correctly:
```C#
var q = Db.From<TestType>().
Where(xx => x.TextCol + x.NullableIntCol.ToString() =="anyString");
```
I implemented "Concat" operator for all database types.


mythz:
Awesome :+1:  you're on a roll :)

OlegNadymov:
Thank you @mythz !

PR #525, opened 2016-07-30
A few minor changes to consolidate SqlExpression.cs

KevinHoward:
A few minor changes to consolidate logic in SqlExpression.cs after the addition of the Concat logic (#a4be4b7).

Thanks @OlegNadymov! Nice addition.


mythz:
good catch, thx Kevin :+1:

PR #527, opened 2016-08-03
Add logic to handle EnumAsInt attribute for Oracle databases

CormacBrennan-AI:
Hi @mythz.

This PR adds logic to handle the EnumAsInt attribute for Oracle databases.

Several of the tests in ServiceStack.OrmLite.Tests.EnumTests which cover this functionality are currently failing on Oracle. They will continue to fail even with these changes as they use hand-coded SQL containing the '@' symbol to specify parameters. Obviously I cannot change this, as the Oracle tests are just links to the main test project. However, when I modify the tests locally to use ':" instead of '@' they do pass on Oracle.

I have signed the CLA.

Perhaps @BruceCowan-AI  would like to take a look at this?


mythz:
ok looks good, thx 👍

PR #528, opened 2016-08-05
Skip InsertOnly tests for Oracle

CormacBrennan-AI:
Hi @mythz,

We run the Oracle tests periodically. There are a few that currently fail. We are in the process of addressing some of the causes of these failures, e.g. #527. However there are some failing tests which will never work on Oracle. We would prefer to just ignore these. Please let me know if that's any kind of problem.

FYI @BruceCowan-AI


mythz:
Yep ignoring is the way to go then.

PR #529, opened 2016-08-08
Add ExecNonQuery overload with dbCmdCallback

georgehemmings:
This provides an alternative to SqlProc for executing non-query stored procedure with output params. This way uses a callback rather than returning a dbCmd (like SqlProc). It can be used the same way SqlList is for queries.


mythz:
no worries, thx George :+1:

PR #537, opened 2016-10-07
Applied some white space changes to the T4 template as the whitespace…

stuartbfs:
… in the generated poco's were slightly off.


mythz:
Great, thx 👍

PR #538, opened 2016-10-13
Changed only accessibiltiy level

OlegNadymov:
Changed accessibiltiy level for method IsColumnAccess


OlegNadymov:
Thanks for merge!

mythz:
np, we're very open to opening things :)

PR #539, opened 2016-10-13
Fix schema related methods in Oracle provider

CormacBrennan-AI:
@mythz - can you take a look please?


mythz:
Looks good from here, thx.

PR #540, opened 2016-10-24
don't need to reset CommandTimeout where it's already set in CreateCommand

georgehemmings:
I've removed the reset of CommandTimeout where it would be set by the CreateCommand in the ExecFilter.

https://forums.servicestack.net/t/setcommandtimeout/3063/4


mythz:
Oh nice I missed those, thx George!

PR #544, opened 2016-11-19
Added MemoryOptimized attribute to SqlServer provider

KevinHoward:
Added a MemoryOptimized attribute to allow for the creation of In-Memory OLTP tables in SQL Server 2014+. This required the the addition of a 2014 and 2016 provider as this syntax wasn't available in 2012.

With this addition, developers will be able to use SQL Server as both a data store and a cache store as illustrated in this article: [https://blogs.msdn.microsoft.com/sqlcat/2016/10/26/how-bwin-is-using-sql-server-2016-in-memory-oltp-to-achieve-unprecedented-performance-and-scale/](https://blogs.msdn.microsoft.com/sqlcat/2016/10/26/how-bwin-is-using-sql-server-2016-in-memory-oltp-to-achieve-unprecedented-performance-and-scale/).

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer/SqlServerOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.SqlServer/SqlServerOrmLiteDialectProvider.cs
@@ -65,16 +65,14 @@ namespace ServiceStack.OrmLite.SqlServer
             {
                 var filePath = connectionString;
 
-                var filePathWithExt = filePath.ToLower().EndsWith(".mdf")
+                var filePathWithExt = filePath.EndsWithIgnoreCase(".mdf")
                     ? filePath
                     : filePath + ".mdf";
 
                 var fileName = Path.GetFileName(filePathWithExt);
                 var dbName = fileName.Substring(0, fileName.Length - ".mdf".Length);
 
-                connectionString = string.Format(
-                @"Data Source=.\SQLEXPRESS;AttachDbFilename={0};Initial Catalog={1};Integrated Security=True;User Instance=True;",
-                    filePathWithExt, dbName);
+                connectionString = $@"Data Source=.\SQLEXPRESS;AttachDbFilename={filePathWithExt};Initial Catalog={dbName};Integrated Security=True;User Instance=True;";
             }
 
             if (options != null)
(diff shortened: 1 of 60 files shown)


src/ServiceStack.OrmLite.SqlServer/SqlServerOrmLiteDialectProvider.cs
```diff
@@ -155,7 +153,7 @@ public override string GetForeignKeyOnDeleteClause(ForeignKeyConstraint foreignK
 
         public override string GetForeignKeyOnUpdateClause(ForeignKeyConstraint foreignKey)
         {
-            return "RESTRICT" == (foreignKey.OnDelete ?? "").ToUpper()
+            return "RESTRICT" == (foreignKey.OnUpdate ?? "").ToUpper()
```
KevinHoward: Looks like GetForeignKeyOnUpdateClause is a copy of GetForeignKeyOnDeleteClause, but was not updated.

mythz:
Fantastic, many thanks for this feature Kevin 👍 

BTW I think I might move the new MemoryOptimized attributes over to ServiceStack.Interfaces along with the new `PgSql*` attributes to avoid a concrete dependency on data models.

KevinHoward:
Sounds good @mythz! Time permitting, I may look at creating a ServiceStack.Caching.SqlServer that uses this new functionality.

KevinHoward:
BTW... I thought about adding the new attribute to ServiceStack.Interfaces, but hesitated because it was something specific to SQL Server. I choose this approach because it's only included when the SqlServer library is referenced, but I probably should have put it in its own file. Same thing is probably the case for your PgSql* attributes.

mythz:
Yeah I had the same hesitation which is why I added the `PgSql*` attributes to the PgSql concrete library, but after thinking about the multiple forces that are in play I've decided to move it to ServiceStack.Interfaces as removing implementation coupling from data annotations justifies introducing provider-specific data attributes which we're kind of already doing for the `[HashKey]` and `[RangeKey]` attributes which are DynamoDB-specific. 

To make it clear that it's a provider-specific attribute I've renamed it to `[SqlServerMemoryOptimized]` to make it clear it's SqlServer specific (the durability makes it unlikely it'll be reusable by any other RDBMS), the `SqlServer*` prefix is good to have on the attribute name as it indicates that the Data Model relies on provider-specific features and will also help with discovery of other SqlServer-specific data annotations in future. These changes is now available from [this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/e1ad7ebbe695b1e6f8e44917fd48e6fd50a8b95e).

mythz:
FYI if you need it this change is [now on MyGet](http://docs.servicestack.net/myget)

PR #548, opened 2016-11-29
Fixed regression caused by last commit and fixed root issue.

KevinHoward:
1. Fixed previously existing issue in construction of Update command where AppendFieldCondition was being called instead of Field = Parameter. (line 767 in OrmLiteDialectProviderBase.cs)
2. Added a new virtual AppendNullFieldCondition method to OrmLiteDialectProviderBase.cs to handle Spatial null conditions (uses a boolean property and not the IS NULL syntax).
3. Added AppendFieldCondition and AppendNullFieldCondition overrides to SqlServer2012OrmLiteDialectProvider.
4. Made dialect provider an optional parameter in the OpenDbConnection method on the SqlServerTests.

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer/SqlServer2012OrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.SqlServer/SqlServer2012OrmLiteDialectProvider.cs
@@ -1,4 +1,6 @@
-﻿using ServiceStack.Text;
+﻿using System.Data;
+using System.Text;
+using ServiceStack.Text;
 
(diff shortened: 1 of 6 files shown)


mythz:
This approach seems ok for now while the exceptional handling is just limited to SQL Server's special types, but if in future we need to do this for more types it would make sense to move this behavior to Type Converters so each Type can define their own custom logic for this.

But it's good for now, thx Kevin!

mythz:
FYI this change is now up on MyGet.

PR #552, opened 2016-12-06
Update to 3.1.8 as fails to find DLL at runtime.

Layoric:
Feel free to ignore PR if incorrect (still find my way around .NET Core).

When trying to use the `ServiceStack.OrmLite.PostgreSQL.Core` package, it fails at runtime due to not being able to find `Npgsql 3.1.8`. This happens for both local Windows dev and running on linux.

If I add the `Npgsql 3.1.8` specifically, it works fine. Possibly cause it's depending on wrong version in nuspec? 

https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/ServiceStack.OrmLite.PostgreSQL/project.json#L20

Or is there another way to get the downstream dependencies to be added to the `project.json`? (this could just be a tooling issue as well I guess).


mythz:
thx, Hopefully this should do it.

PR #555, opened 2016-12-20
Add update with db supplied timestamp

BruceCowan-AI:
@mythz You may remember that a relatively long time ago I added an `OnUpdate` operand to the `Default` attribute, the idea being that when `OnUpdate` was true, the default value would be supplied automatically when doing an update operation. The main use case is to have the database update a lastupdated timestamp field to the database time when you do an update. Well, I finally got around to implementing it, and here is a pull request for the changes. The code supplies the default value for any column that has a `Default` attribute with `OnUpdate` true, because it doesn't know anything about timestamps, of course.

I also fixed a few Oracle test failures, mostly with changes to the tests to handle Oracle differences.

I've run SQLite, SQL Server, and Oracle tests and it all seems good to me. Sorry, I don't have time to try out the other databases.

Be aware that I made one non-backward-compatible changes to the `IOrmLiteDialectProvider` interface - I removed the unused `updateFields` parameter from `PrepareParameterizedUpdateStatement` and I added a return value to it.

Hopefully this is all good with you. If not, ask and I will fix it (we really need the ability to put database timestamps in on updates).

### DIFF
--- a/src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
@@ -624,34 +624,47 @@ namespace ServiceStack.OrmLite.Oracle
             }
         }
 
-        // TODO: do we need this?
-        public override void SetParameterValue<T>(FieldDefinition fieldDef, IDataParameter p, object obj)
+        protected override void SetParameterValue<T>(IDbCommand dbCmd, ModelDefinition modelDef, Dictionary<string, FieldDefinition> fieldMap, IDataParameter p, object obj)
         {
-            p.Value = GetValueOrDbNull<T>(fieldDef, obj);
+            FieldDefinition fieldDef;
+            var fieldName = this.ToFieldName(p.ParameterName);
+            fieldMap.TryGetValue(fieldName, out fieldDef);
+
+            if (fieldDef == null)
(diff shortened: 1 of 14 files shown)


BruceCowan-AI:
@mythz If you have any problems with this PR, please just tell me and I'll happily rework it and refactor the code as necessary. I think I'll split off the fixing of tests for Oracle from the main commit (6843148). Thanks, Bruce

mythz:
Well my initial thoughts is that I'm concerned about the complexity/benefit trade-off of this feature which is too disruptive/breaking for an unknown feature that's unlikely to be used by anyone else. Not sure if it's possible atm, but I'd prefer to expose hooks instead of baking in the implementation into OrmLite.

Could you separate the changes to with the `[Default(OnUpdate=true)]` feature in a separate PR (inc tests) so I can merge the other changes first and have a better idea of the changes required to support this feature independently?

BruceCowan-AI:
No problem, will separate the other changes into a separate PR.

I'll think about the complexity/benefit trade-off. IMO, anyone who has multiple app machines ought to be using database timestamps for any current time in the database, so they avoid discrepancies from unsynchronized clocks, so I would expect that it is actually more applicable than you say.

BruceCowan-AI:
Closing this PR since it is no longer correct and github didn't remove the commits I don't want in here, even though they are not in my branch any more.

PR #556, opened 2016-12-21
Fix oracle provider testing somewhat

BruceCowan-AI:
@mythz Here is a small PR which fixes a bunch of tests on the Oracle provider. Mostly it changes the tests, not the provider. I verified that the changes don't break other providers by running the tests on SQL Server.


mythz:
thx Bruce 👍

PR #557, opened 2016-12-22
Add support for [Default(OnUpdate=true)]

BruceCowan-AI:
@mythz here is the OnUpdate pull request again, with no extraneous stuff in it. I did some refactoring to remove duplication and speed up the sql generation. I would really appreciate if you would consider this as we really need it in quite a number of places in our product and really want to avoid bypassing OrmLite in those places. As I said in an earlier message today, the ability to set a column to a database-supplied date on update is really very useful and I'm sure others will use it, once they see that the capability exists in OrmLite in an easy to use way.

However, I'd be happy to discuss this, and I'd very much appreciate doing that quickly as I have vacation planned from Friday through to after New Years.

Thanks, Bruce

### DIFF
--- a/src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.Oracle/OracleOrmLiteDialectProvider.cs
@@ -377,6 +377,48 @@ namespace ServiceStack.OrmLite.Oracle
             }
         }
 
+        protected override void SetParameterValue<T>(IDbCommand dbCmd, ModelDefinition modelDef, Dictionary<string, FieldDefinition> fieldMap, IDataParameter p, object obj)
+        {
+            FieldDefinition fieldDef;
+            var fieldName = this.ToFieldName(p.ParameterName);
+            fieldMap.TryGetValue(fieldName, out fieldDef);
+
+            if (fieldDef == null)
+                throw new ArgumentException("Field Definition '{0}' was not found".Fmt(fieldName));
+
+            if (fieldDef.AutoIncrement || !string.IsNullOrEmpty(fieldDef.Sequence))
+            {
+                if (fieldDef.AutoIncrement && string.IsNullOrEmpty(fieldDef.Sequence))
+                {
+                    fieldDef.Sequence = Sequence(NamingStrategy.GetTableName(modelDef), fieldDef.FieldName, fieldDef.Sequence);
+                }
+
+                var pi = typeof(T).GetProperty(fieldDef.Name,
+                    BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.FlattenHierarchy);
+
+                //TODO fix this hack! Because of the way we handle sequences we have to know whether
+                // this is an insert or update/delete. If we did sequences with triggers this would
+                // not be a problem.
+                var sql = dbCmd.CommandText.TrimStart().ToUpperInvariant();
+                bool isInsert = sql.StartsWith("INSERT", StringComparison.InvariantCulture);
+
+                var result = GetNextValue(dbCmd.Connection, dbCmd.Transaction, fieldDef.Sequence,
+                                          pi.GetValue(obj, new object[] { }), isInsert);
+                if (pi.PropertyType == typeof(String))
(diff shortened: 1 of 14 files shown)


mythz:
Hi Bruce,

I appreciate the amount of effort, but it's still a breaking change and just too invasive touching so many source files with tight implementation coupling leaking into multiple code paths throughout OrmLite's code-base for such a small feature that's never been requested before. It's already hard to reason about in isolation within a single PR, if it were to be integrated with the code-base it'd be much worse. The complexity cost just isn't justifiable for this feature which would hinder refactoring and make it harder for others to understand the code-base. Overall I see this PR as a net-negative.

The `[Default(1, OnUpdate=true)]` also has an uncommon UX which I'm not excited about, most of OrmLite's attributes are declarative or follows a close mapping with SQL - although I'm not sure what it should be.

It sounds like you're in a rush, but I can't see myself accepting the PR as it is, once its shipped we'd have to live with this complexity forever which is contra to our core objectives of enabling maximum value with minimal complexity. But if you need this in short-term I think you'd have to consider using a custom build of OrmLite for the time being. I'll look to see if there's a nicer way of adding `the ability to set a column to a database-supplied date on update` which I think is the only useful use-case.

BruceCowan-AI:
Well, you did accept `[Default(1, OnUpdate=true)]` quite some time ago - in fact I think it was  and we did discuss this then. We also discussed other ways of adding the ability to set a column to a database-supplied date at that time and you were dead set against anything other than `[Default]`. I had proposed another way and you rejected it at that time.

I agree that the complexity cost is rather high. But it isn't a breaking change in that it doesn't break any of the public API. I also agree that the only useful use case is setting a column to a database supplied date on update.

BruceCowan-AI:
Hi Demis,

I understand that no one but us has requested this capability, but in a scale-out world, using the database time for timestamping updates is really the only correct way to do it.

I'd really like to resolve the UX asap. As a suggestion for discussion, how about:

- remove the `OnUpdate` argument from the `Default` attribute. This is a breaking change, but since nothing implemented it, no one should be using it.
- add new `Update` and friends methods overloads which take an additional parameter `Expression<Func<T>> defaultTimestamps` to specify a timestamp field which has a `Default` attribute and which has `DateTime` type (although using an expression means someone could specify more than one field), _OR_
- add new `Update` and friends method overloads which take an additional parameter `string timestampField` which must specify a field which has a `Default` attribute and which has `DateTime` type.

I think I prefer the latter version, the one with a field name because you can use it via `nameof(poco.FieldName)` and because allowing more than one timestamp field to be updated is very strange.

What say you?

Thanks,
Bruce

BruceCowan-AI:
Hi Demis,

I'm available to discuss this tomorrow (Friday), even though I'm on vacation. Could even discuss by phone if that would work.

A few more notes about the above proposal:
- Possibly customers aren't asking for this because they saw it missing and decided against OrmLite as a result.
- The `Update` and friends methods really ought to be simplified. There is too much overlap.
- Because I'm proposing adding more update extension methods, it should be easy to limit the number of files changed.

mythz:
Hi Bruce, before trying writing any code to see if it would actually work, my initial thoughts was to investigate adding an Update overload that accepts a callback. But I was hoping for something more generic/useful like a `Func<IDbCommand>` callback where you'd be able to modify the populated dbCmd/parameters before it's executed.

BruceCowan-AI:
Hi Demis, I guess I could make that work for me, but a bit awkward because we support two different databases as a configuration option in our product. I still think that is not the best choice for this capability because it effectively hides it, and as I've said, I think it is a capability many others would use if it was offered.

mythz:
I'd prefer to make it possible under a generic API first before committing to what the optimal way to expose this feature is. Adding a specific overloads like `Expression<Func<T>> defaultTimestamps` for every minor feature unnecessarily complicates the API Surface area. Having it working on top of a generic API allows others to add a more specific reusable API in their code-base via a UX-friendly extension method without having to bake a concrete implementation in OrmLite's internals.

BruceCowan-AI:
I understand. It is important that the generic API include the poco mapped to the table. Do you want me to implement such an API?

mythz:
Yes please.

PR #558, opened 2016-12-22
Fix ToUpdateStatement and ToInsertStatement with subset named columns

BruceCowan-AI:
@mythz Here is a simple little fix to `ToUpdateStatement` and `ToInsertStatement` which fail when one field in the poco has a name which is a leading subset of another field. I ran into it with my update changes. I think it is safe enough as field names should not contain special regular expression characters. Cheers, Bruce


mythz:
thx Bruce.

PR #559, opened 2016-12-23
A start at implementing generic update callback

BruceCowan-AI:
@mythz THIS IS NOT READY FOR MERGE, here for your comments only. I think this generic update callback is close to what you wanted, and the test I added to `DefaultValueTests` shows using it to insert a last updated date (once I fix the test to do the substitution into the sql correctly). Obviously it also needs more tests. Please comment at will.

Cheers, Bruce

### DIFF
--- a/src/ServiceStack.OrmLite/Async/OrmLiteWriteCommandExtensionsAsync.cs
+++ b/src/ServiceStack.OrmLite/Async/OrmLiteWriteCommandExtensionsAsync.cs
@@ -52,10 +52,9 @@ namespace ServiceStack.OrmLite
             return dbCmd.GetDialectProvider().ExecuteNonQueryAsync(dbCmd, token);
         }
 
-        internal static Task<int> UpdateAsync<T>(this IDbCommand dbCmd, T obj, CancellationToken token)
+        internal static Task<int> UpdateAsync<T>(this IDbCommand dbCmd, T obj, CancellationToken token, Action<IDbCommand> commandFilter)
         {
-            if (OrmLiteConfig.UpdateFilter != null)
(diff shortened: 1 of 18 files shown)


mythz:
Yep looks good, I'm much happier with this approach, thx.

BruceCowan-AI:
@mythz I think this is complete now and ready to merge. There are tests for every change I made. Perhaps I put the async tests in the wrong place, as I noticed that some other async tests seem to be in provider-specific files and I didn't do that. I ran the tests only on Oracle and SQL Server.

Cheers, Bruce

src/ServiceStack.OrmLite/Async/OrmLiteWriteCommandExtensionsAsync.cs
```diff
@@ -73,12 +73,12 @@ internal static Task<int> UpdateAsync<T>(this IDbCommand dbCmd, T obj, Cancellat
                 });
         }
 
-        internal static Task<int> UpdateAsync<T>(this IDbCommand dbCmd, CancellationToken token, params T[] objs)
+        internal static Task<int> UpdateAsync<T>(this IDbCommand dbCmd, CancellationToken token, Action<IDbCommand> commandFilter, T[] objs)
         {
```
mythz: seems like we missed a `params` here?

BruceCowan-AI: I removed the `params` because this is an internal method which cannot be called from anything other than some other part of OrmLite, so `params` isn't needed.

mythz:
Everything looks good from here, thx Bruce 👍 

FYI this change is now available from v4.5.5+ on MyGet.

PR #562, opened 2017-01-04
[OrmLite.SqlServer] Added COLLATE Column option, Memory-Optimized Table Option, & FileTable Table Option with Unit Tests

KevinHoward:
1. Updated CreateTable methods for SqlServer 2012+ Dialect Provider to construct FileTable table type (new sbTableOptions stringbuilder).
2. Updated GetColumnDefinition method with COLLATE Column option (only affects string types, i.e. VARCHAR, NVARCHAR, etc.).
3. Updated GetColumnDefinition method with In-Memory Optimized BUCKET COUNT & NONCLUSTERED PRIMARY KEY (Primary Key can't be clustered for this table type) options.
4. Updated CreateTable methods for SqlServer 2014+ Dialect Provider to construct In-Memory Optimized.
5. Updated AppendFieldCondition method to handle comparing SqlServer Spatial Types (STEquals method).
6. Updated AppendNullFieldCondition  method to handle null checking for SqlServer Spatial & HierarchyId Types (IsNull method).
7. Provided Unit Testing that illustrates the Memory-Optimized (with different Durabilities) and FileTable features, plus the use of the COLLATE column option. 
8. Other misc. fixes such as C# 6-ifying a few existing methods and removing unused using statements.

TODO: Move SqlServerFileTableAttribute to ServiceStack.Interfaces along side SqlServerMemoryOptimizedTableAttribute.


mythz:
Great thx Kevin 👍 

FYI `SqlServerFileTableAttribute` has now been moved to SS.Interfaces and is now available under `/lib`.

PR #564, opened 2017-01-26
Fix the default value of the "token" parameter in Async methods

shift-evgeny:
Fix the default value of the "token" parameter in Async methods of classes derived from OrmLiteDialectProviderBase to match those in the base class. Without this calling the methods via an instance of the derived class requires specifying a CancellationToken.


mythz:
ok thanks, we don't consider these async methods as external user-facing methods which is why they didn't have optional Cancellation Tokens before, but cool to have them here as well if they're useful to you.

PR #568, opened 2017-02-17
Change access modifier for InternalJoin to protected and virtual

OlegNadymov:


OlegNadymov:
@mythz thanks for merging! :+1:

PR #569, opened 2017-02-19
Use GetQuotedColumnName instead of DialectProvider.GetQuotedColumnName inside SqlExpression

OlegNadymov:
Use GetQuotedColumnName instead of DialectProvider.GetQuotedColumnName inside SqlExpression.

Method GetQuotedColumnName in SqlExpression is virtual. Therefore this refactoring will add additional features for SqlExpression inheritors.

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -137,6 +137,8 @@ namespace ServiceStack.OrmLite
             if (fields == null || fields.Length == 0)
                 return Select(string.Empty);
 
+            useFieldName = true;
+
             var allTableDefs = new List<ModelDefinition> { modelDef };
             allTableDefs.AddRange(tableDefs);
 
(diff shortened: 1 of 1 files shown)


OlegNadymov:
Could you show more details about error?

mythz:
This test is failing:
```
PrimaryExpressionsTest.Can_select_bool_property_expression 	
First failure:	569/head	#1654 	olegnadymov (1) 	19 Feb 17 16:57
   Expected: greater than 1
  But was:  1

at ServiceStack.OrmLite.Tests.Expression.PrimaryExpressionsTest.Can_select_bool_property_expression() in C:\TeamCity\buildAgent\work\12884bd5feef0ce7\tests\ServiceStack.OrmLite.Tests\Expression\PrimaryExpressionsTest.cs:line 308
```

Do you have a test showing why you need this change?

OlegNadymov:
I have passed this test on my local machine.
I don't understand why teamcity has fail there.

OlegNadymov:
I have custom logic in GetQuotedColumnName in SqlExpression inheritor. 
I try use OrmLite for classes with hierarchy (the data is stored in different tables).

If I need to create a test for that I'll try to make it.

mythz:
Seems like the test was flaky, it passed after running it again.

A test on why you needed this change will prevent a regression when this code is refactored. i.e. If there's no test that explains why this change is needed it allows a future refactor to remove it.

OlegNadymov:
Ok! I've understood about tests. I'll think about test for this changes.
Thank you!

OlegNadymov:
I've added unit test for that commit:
https://github.com/ServiceStack/ServiceStack.OrmLite/pull/571

PR #576, opened 2017-04-04
Fixed class name in reference attribute

englerj:
Fixed an issue where the reference attribute would ignore the class prefix and suffix, causing a 'Cannot resolve symbol' error when IncludeReferences is true.


mythz:
great, thx

PR #580, opened 2017-05-11
SqlVerifyFragment improvement.

Naragato:
The current method of checking leads to too many false positives, also having it iterate over each key + do additional operations to prevent this is possibly performs worse than a regex (this would need to be checked properly ofc).


mythz:
Thanks for the PR! The compiled regex should be static, but I can fix that on my side. Can you please [approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so we can merge this, thx.

Naragato:
Oops I thought I already did !

mythz:
Cheers 👍

mythz:
FYI I made it static and added an ability to easily overload the impl in [this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/68de00e47329b047f15da0f97f789b43e505cc03).

This change is available from v4.5.9 that's now [available on MyGet](http://docs.servicestack.net/myget).

Naragato:
Nice :D

PR #581, opened 2017-05-22
Change access modifiers only

OlegNadymov:


mythz:
ok thanks but I'm not a fan of exposing internal APIs like Visit to the public end-user API, but I take it you need access to it

OlegNadymov:
Thank you!
Public ``Visit()`` is like public ``Visit()`` in ``System.Linq.Expressions.ExpressionVisitor``. It allows to write extensions for ``SqlExpression``, e.g. move Sql-code creation to external extensions of such methods like ``Sql`` methods or other custom methods. Theese extensions need public ``Visit()`` method.

OlegNadymov:
Thank you!
Public ``Visit()`` is like public ``Visit()`` in ``System.Linq.Expressions.ExpressionVisitor``. It allows to write extensions for ``SqlExpression``, e.g. move Sql-code creation to external extensions of such methods like ``Sql`` methods or other custom methods. Theese extensions need public ``Visit()`` method.

PR #583, opened 2017-06-07
Changed the ParamNameFilter to a instance variable on the Dialect Provider

stuartbfs:
This is to support being able to use ormlite in an application querying both sql server and oracle databases where the Oracle data provider is Devart.

Currently to use OrmLite with Devart, we need to set:
``
OrmLiteConfig.ParamNameFilter = paramName => $":{paramName}"
``
But because OrmLiteConfig is static, this breaks a separate Sql Server connection running in the same application.

This change would allow the ParamNameFilter to be overwritten for a specific dialect provider, if required.


mythz:
Cheers 👍

PR #585, opened 2017-06-26
Update SqliteOrmLiteDialectProviderBase.cs

alpha-before-beta:
When using the in memory db for Sqlite, schemas with a period cause the object to not be found as cross database operations are not supported. This change will prevent this error by injecting an underscore. Similar to the implementation found in GetTableName.


alpha-before-beta:
Closing this. To be deleted

PR #586, opened 2017-06-26
Update OrmLiteDialectProviderBase.cs, SqliteOrmLiteDialectProviderBase.cs

alpha-before-beta:
Added virtual method GetSchemaName
Added override for sqlite implentation
Made these changes to account for cross database operations that use a period in the schema name.

### DIFF
--- a/src/ServiceStack.OrmLite.Sqlite/SqliteOrmLiteDialectProviderBase.cs
+++ b/src/ServiceStack.OrmLite.Sqlite/SqliteOrmLiteDialectProviderBase.cs
@@ -140,6 +140,13 @@ namespace ServiceStack.OrmLite.Sqlite
 
         protected abstract IDbConnection CreateConnection(string connectionString);
 
+        public override string GetSchemaName(string schema)
+        {
+            return schema != null
+                ? NamingStrategy.GetSchemaName(schema).Replace(".", "_")
+                : NamingStrategy.GetSchemaName(schema);
+        }
+
         public override string GetTableName(string table, string schema=null)
         {
             return schema != null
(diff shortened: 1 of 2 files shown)


mythz:
Please provide an example of why this change is needed, can you also [approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform), thx.

alpha-before-beta:
1. I approved the contributor license agreement

2. This change is needed to address the following scenario...

To start, I am coming from the context of SQL Server. Say we have two databases. Database A and Database B. Both databases contain a schema named MySchema. In database A is a table named Customer. In database B a table named Household. When needing to do a cross database join between these two, one of the DTO's must contain the database name as part of the schema. This results in the period character in the name which in turn Sqlite does not handle. Illustration below. Please notice the schema on the household class.

//table lives in database A
[Schema("MySchema")]
public class Customer {
     public int CustomerId { get; set; }
}

//table lives in database B
[Schema("B.MySchema")]
public class Household {
     public int CustomerId { get; set; }
     pubilc int HouseholdId { get; set; }
}

using (var db = DbFactory.Open("A"))
{
     var query = db.From<Customer>().LeftJoin<Household>((c, h) => c.CustomerId == h.CustomerId);
     return db.Single<FullCustomerInfo>(query);
}

mythz:
BTW Thanks for the PR, FYI if you need it this change is available from v4.5.13 that's now [available on MyGet](http://docs.servicestack.net/myget).

alpha-before-beta:
You're welcome. Thank you for an awesome piece a software. Keep up the great work!

PR #588, opened 2017-07-17
use t4 for maradb...  

hhq365:
when i use ServiceStack.OrmLite.MySql for maradb,the template file OrmLite.Poco.tt will generate repetition attribute for table 'User' and 'Role' and so on.One is mine,another from system table :'performance_schema'... How can i do?


mythz:
We don't support Maria DB in T4 templates, You'll need to manually remove the duplicate tables.

PR #589, opened 2017-07-24
Allow ORMLite Extensions to be used without global dialect provider.

sl33kr:
My team at the company I work for are in the process of migrating from one DBMS to another. In this situation we can't set the global dialect provider as it causes syntax errors across the DBMSs.

This pull request fixes the issue for us for SQL Server, unless there is something we are doing wrong?

When attempting to use ORMLite extensions against and OrmLiteConnection an ArgumentNullException was thrown when calling to SqlFmt as the global provider was not set.
Using SqlFmt overload that accepts the current dialect provider stops the ArgumentNullException.


sl33kr:
Sorry I probably should have also mentioned that we have Business licenses.

mythz:
Great thx for the PR, can you also [approve the Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSc-vKvdOx5lXGqmy6fkdCj-N0pALkVCx7QHYXojAuDSv_enNA/viewform) so we can merge this, thx.

sl33kr:
Thanks @mythz,
Agreement has been submitted.

mythz:
FYI this change is now [available on MyGet](http://docs.servicestack.net/myget)

PR #590, opened 2017-07-26
Added 3 JSON Sql methods to the SQLServer 2016 Dialect Provider

KevinHoward:
Added a few of the JSON SQL methods to the SqlServer 2016 Dialect Provider so developers can parse and query JSON objects and arrays directly from SQL Server. This should be helpful for developers who dump raw JSON into a single string field (NVARCHAR) and still need the ability to query this data. 

For example, if a developer stores an entire address as JSON into a single field and need to only query those where State = "NY", this should make it easier and much more efficient.

Note that my NUnit test adapter has been malfunctioning and I wasn't able to execute the tests, so you may find a few errors. Sorry ;-(.

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
+++ b/src/ServiceStack.OrmLite.SqlServer.Converters/ServiceStack.OrmLite.SqlServer.Converters.csproj
@@ -29,7 +29,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
-    <PackageReference Include="Microsoft.SqlServer.Types" Version="11.0.2" />
+    <PackageReference Include="Microsoft.SqlServer.Types" Version="14.0.314.76" />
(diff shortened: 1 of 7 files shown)


mythz:
Looks great, thx Kevin 👍 I might rework it a bit, i.e. I don't think we need a separate `Sql2016` class as it hurts discoverability and force users having to jump back/forward between `Sql` and `Sql2016`. Users will just need to be aware which features are available in their RDBMS.

mythz:
Umm these tests seem way off. I've reworked this PR in [this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/c21bed6724de654248523609d095602706023ce0) and have refactored the tests to show the types, can you look at making them pass please?

KevinHoward:
Sure thing Demis... trying to figure out why Visual Studio isn't discovering any of the unit tests even though I have the needed adapter installed. May need to do some reinstalling.

As far as the Sql2016 class, you are correct. Didn't think about making Sql partial, but that is probably the best approach.

Not sure if you have played with this functionality on SQL Server or read anything about it, but it's pretty cool.

PR #593, opened 2017-08-17
Update Unit Tests and made some changes around the SQL Server 2016 JSON functions

KevinHoward:
Sorry for the delay on this effort Demis! Had some changes at the paying job and couldn't make it back around until now.

Anyhow, I have gone back through my code and made some changes to get the Unit Tests executing correctly.  I only have one failing case, and it's due to a conversion error (object must inherit IConvertible, not sure why yet), but it's for a scenario that few would probably use.

Take a look and let me know what you think.

Oh, one side note... I found out that I had to add a Microsoft.NET.Test.Sdk NuGet dependency (only to the testing project) so that Visual Studio 2017 can properly discover the NUnit test cases using the proper adapter.


KevinHoward:
Related to initial changes here #590 .

mythz:
Looks good, thx Kevin!

PR #595, opened 2017-08-25
Fix Sql.In condition don't support skip and take by mysql.

wuyou331:


mythz:
Can you please provide a test asserting why this PR is needed, also please don't add provider-specific checks in `SqlExpression`, make a method and override it in `MySqlDialectProvider` instead.

mythz:
Cheers 👍

wuyou331:
When will the new version be released to nuget?

mythz:
We're aiming for within a week, this [change is available on MyGet now](http://docs.servicestack.net/myget) if you need it.

PR #597, opened 2017-09-11
Added possibilities for custom sql expression:

OlegNadymov:
1. ordering by constant
 2. using constant as column access.

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -672,9 +672,13 @@ namespace ServiceStack.OrmLite
             sep = string.Empty;
             useFieldName = true;
             orderByProperties.Clear();
-            var fields = Visit(keySelector).ToString();
-            orderByProperties.Add(fields);
-            BuildOrderByClauseInternal();
+            var orderBySql = Visit(keySelector);
+            if (orderBySql is PartialSqlString)
+            {
+                var fields = orderBySql.ToString();
+                orderByProperties.Add(fields);
+                BuildOrderByClauseInternal();
+            }
             return this;
         }
 
@@ -700,9 +704,13 @@ namespace ServiceStack.OrmLite
         {
             sep = string.Empty;
             useFieldName = true;
-            var fields = Visit(keySelector).ToString();
-            orderByProperties.Add(fields);
-            BuildOrderByClauseInternal();
+            var orderBySql = Visit(keySelector);
+            if (orderBySql is PartialSqlString)
+            {
+                var fields = orderBySql.ToString();
+                orderByProperties.Add(fields);
+                BuildOrderByClauseInternal();
+            }
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


OlegNadymov:
Could you give more details about failed tests, please?

OlegNadymov:
@mythz could you help me ?

OlegNadymov:
I've found problem. I'm working to fix it now.

OlegNadymov:
I've fixed my bug in the PR.

OlegNadymov:
Is something wrong with my PR?

src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
```diff
@@ -682,6 +683,12 @@ private SqlExpression<T> OrderByInternal(Expression keySelector)
             return this;
         }
 
+        private static bool ObjectIsClass(object obj)
+        {
+            return obj != null &&
+                   !(obj.GetType().IsPrimitive() || obj.GetType().IsValueType() || obj is string);
+        }
+
```
mythz: Why isn't this using `ob.GetType().IsClass()`?

OlegNadymov: I need to determine value types include ``string``. The ``ob.GetType().IsClass()`` is ``true`` for ``string``. Also I've found solution for this problem in your project in method ``ServiceStack.OrmLite.OrmLiteSPStatement.ConvertToScalarList<T>()``:
```
public List<T> ConvertToScalarList<T>()
        {
            if (!((typeof(T).IsPrimitive()) || typeof(T).IsValueType() || (typeof(T) == typeof(string)) || (typeof(T) == typeof(String))))
                throw new Exception("Type " + typeof(T).Name + " is a non primitive type. Use ConvertToList function.");
...
        }
```
So I can change like this:
```
        private static bool ObjectIsClass(object obj)
        {
            return obj != null &&
                   !(obj is string) &&
                   obj.GetType().IsClass();
        }
```
Is it necessary to do so?

mythz: IMO It makes more sense for the method called `ObjectIsClass()` to use the `IsClass()` API then using inverted checks of IsPrimitive + IsValueType.

But I'm not thrilled of the name of the method as it doesn't provide any indication of what it's guarding against, what other types is this method for apart from `PartialSqlString` ?

OlegNadymov: One more type is ``SelectList`` (``VisitNew(NewExpression nex``). So i've decided exclude all reference types.

mythz: It's more descriptive to list the concrete types so we know when this code-path is used.

OlegNadymov: Something like this?
```
        private static bool ObjectIsSqlClass(object obj)
        {
            return obj != null &&
                   (obj is PartialSqlString ||
                    obj is SelectList);
        }
```

mythz: Just IsSqlClass is fine.

OlegNadymov:
Ok! I've done it.

mythz:
Great, thx 👍

OlegNadymov:
@mythz  Thanks for your advice!

PR #598, opened 2017-09-14
Expanding possibilities for custom sql expressions

OlegNadymov:
1. Removed checking by string for using parameters
2. Using constant in select part


mythz:
Cool, thx 👍

OlegNadymov:
@mythz thanks! 👍

PR #601, opened 2017-10-11
Add support of SQL-generation for String.Compare

OlegNadymov:
I have one question.
@mythz How can I import key "servicestack-sn.pfx" which protected with password ?


mythz:
From now onwards we're signing all projects so I checked in my pfx signing key, unfortunately I forgot it was password protected and VS.NET doesn't seem to allow using an unprotected pfx key so I've had to switch to use servicestack.snk which is not password protected so you should now be able to build all projects.

BTW thx for the PR 👍

OlegNadymov:
I've understood you. Thank you!

PR #602, opened 2017-10-20
My sql connector

Naragato:
Creating a dialect provider for https://mysql-net.github.io/MySqlConnector/.  Most of the logic is compatible with the official MySQL connector.  The commonality has been moved shared projects.   This initial pull request is to open discussion for changes / suggestions.


mythz:
ok great thx for the massive PR! We'll check it out to see if it passes the test suite, if it does we'll include it in our CI/deployments to NuGet.

mythz:
FYI we don't use shared projects, I've converted this PR back to use normal multi-platform projects that uses source-code linking to include the same shared classes in multiple projects.

The `ServiceStack.OrmLite.MySqlConnector.Tests` project is just a [.csproj in ServiceStack.OrmLite.MySql.Tests](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/ServiceStack.OrmLite.MySql.Tests/ServiceStack.OrmLite.MySqlConnector.Tests.csproj) so it uses the same source files and a `MYSQLCONNECTOR` build symbol to distinguish between the 2 MySql projects.

VS.NET msbuild's project system doesn't seem to support this scenario too well as a rebuild of the entire solution yields:

```
'C:\src\ServiceStack.OrmLite\src\ServiceStack.OrmLite.MySql\ServiceStack.OrmLite.MySql.csproj'. This can indicate a missing project reference.	ServiceStack.OrmLite.MySqlConnector.Tests	C:\Program Files\dotnet\sdk\2.0.0\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.Sdk.targets	114	
```

Which is looking for the `ServiceStack.OrmLite.MySql.csproj` project in the `ServiceStack.OrmLite.MySqlConnector.Tests` project which doesn't exit, this is likely a result from using the same `/obj` working folder and not being able to support multiple .csproj's in the same folder so I've disabled building the project from the default build configurations. To run the tests you'll need manually build the `ServiceStack.OrmLite.MySqlConnector.Tests` project, then you can run the tests.

The other difference with https://github.com/mysql-net/MySqlConnector is that it isn't signed so we can't sign `ServiceStack.OrmLite.MySqlConnector` which is unfortunate as [all our other projects are now Signed](https://forums.servicestack.net/t/myget-pre-release-packages-upgraded-to-v5/4749).

Anyway thanks for the PR, I've added a `ServiceStack.OrmLite.MySqlConnector` NuGet package for it which is available from `v5.0.0` that's now [available on MyGet](http://docs.servicestack.net/myget) if you need it. Please note the [other v5 changes](https://forums.servicestack.net/t/myget-pre-release-packages-upgraded-to-v5/4749) before upgrading.

Naragato:
@mythz looks good, thanks :) I'll try to get the MySqlConnector project to look into signing their packages.

bgrainger:
> The other difference with https://github.com/mysql-net/MySqlConnector is that it isn't signed so we can't sign ServiceStack.OrmLite.MySqlConnector

@mythz I've fixed this in MySqlConnector 0.30.0.

mythz:
@bgrainger Awesome, new release of ServiceStack.OrmLite.MySqlConnector on MyGet is now signed, Cheers 👍

PR #603, opened 2017-11-26
Feature/rowversion: make RowVersion logic cater for ulong and byte[] properties

daleholborow:
Follow up to https://forums.servicestack.net/t/cant-use-ormlite-rowversion-and-dapper-custom-sql-in-same-project/4947/4 in which I proposed allowing byte[] properties as RowVersion trackers so that the Dapper functionality embedded in ServiceStack can be used with records containing RowVersions (ulong is inconsistent with Dapper mapping and so crashes when using the two in combination).

All tests implemented for this new functionality are in the RowVersionTests.cs file.

Broadly speaking, the RowVersionConverter has become aware of the tracking field type within the FromDbValue method so that it can decide whether to force a parse to ulong or byte[] , and the FromDbRowVersion method was subsequently made redundant and removed.

The solution builds, BUT i do not have access to the db types other than MS SQL nor could I run those test cases for those other db types - if you can run those tests and let me know if there are any issues, that would be great.

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer/Converters/SqlServerSpecialConverters.cs
+++ b/src/ServiceStack.OrmLite.SqlServer/Converters/SqlServerSpecialConverters.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Data;
 using ServiceStack.OrmLite.Converters;
 
 namespace ServiceStack.OrmLite.SqlServer.Converters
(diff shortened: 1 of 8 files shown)


mythz:
Great thanks, the other DB's just needed this commit https://github.com/ServiceStack/ServiceStack.OrmLite/commit/9077c590828e5a31cff91496e7e0b2764ade0427 to pass, now all's good.

Cheers 👍

PR #606, opened 2018-02-20
Implements the ReturnOnInsert attribute

madaleno:
Implements the ReturnOnInsert attribute to return field values in Insert statements.
This solves the problem of using Sequences instead of Identity for AutoIncrement fields.
Also can be used to returns fields that have their value changed/calculated by triggers after the insert.

Works well in SqlServer.
Does not work with FireBird yet. Probably the FireBird .net client does not support the returning clause.

The code still needs some revision and optimization.
Will add tests in a next commit.

Defining this code:

```C#
    [Alias("Users")]
    public class User
    {
        [ReturnOnInsert]
        [PrimaryKey]
        [AutoIncrement]
        public int Id { get; set; }
        string Name  { get; set; }
        string Email  { get; set; } 
        [ReturnOnInsert]
        string Country  { get; set; } 
     }
```  
```C#
var user = new User { Name = "Me", Email = "me@me.com" }
db.Insert(user);
// using SqlServer, user.Id now has the value of the database field even not using Identity.
```

### DIFF
--- a/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
@@ -206,7 +206,7 @@ namespace ServiceStack.OrmLite.Firebird
                     sbReturningColumns.Append(GetQuotedColumnName(fieldDef.FieldName));
                 }
 
-                if (fieldDef.ShouldSkipInsert() && !fieldDef.AutoIncrement)
+                if (fieldDef.ShouldSkipInsert() && !fieldDef.AutoIncrement && string.IsNullOrEmpty(fieldDef.Sequence))
                     continue;
 
                 //insertFields contains Property "Name" of fields to insert ( that's how expressions work )
@@ -222,15 +222,15 @@ namespace ServiceStack.OrmLite.Firebird
                 {
                     sbColumnNames.Append(GetQuotedColumnName(fieldDef.FieldName));
 
-                    if (!fieldDef.AutoIncrement)
-                    {
-                        sbColumnValues.Append(this.GetParam(SanitizeFieldNameForParamName(fieldDef.FieldName)));
-                        AddParameter(cmd, fieldDef);
-                    }
-                    else
+                    if (fieldDef.AutoIncrement || !string.IsNullOrEmpty(fieldDef.Sequence))
                     {
                         EnsureAutoIncrementSequence(modelDef, fieldDef);
                         sbColumnValues.Append("NEXT VALUE FOR " + fieldDef.Sequence);
+                    }
+                    else
+                    {
+                        sbColumnValues.Append(this.GetParam(SanitizeFieldNameForParamName(fieldDef.FieldName)));
+                        AddParameter(cmd, fieldDef);
                     }
                 }
                 catch (Exception ex)
--- a/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
+++ b/src/ServiceStack.OrmLite.Firebird/FirebirdOrmLiteDialectProvider.cs
@@ -118,12 +118,19 @@ namespace ServiceStack.OrmLite.Firebird
 
             var sbColumnNames = StringBuilderCache.Allocate();
             var sbColumnValues = StringBuilderCacheAlt.Allocate();
+            var sbReturningColumns = StringBuilderCacheAlt.Allocate();
 
             var tableType = objWithProperties.GetType();
             var modelDef = GetModel(tableType);
 
             foreach (var fieldDef in modelDef.FieldDefinitionsArray)
             {
+                if (fieldDef.ReturnOnInsert || (fieldDef.IsPrimaryKey && fieldDef.AutoIncrement && modelDef.HasReturnAttribute))
+                {
+                    if (sbReturningColumns.Length > 0)
+                        sbReturningColumns.Append(",");
+                    sbReturningColumns.Append(GetQuotedColumnName(fieldDef.FieldName));
+                }
 
                 if (fieldDef.IsComputed)
                     continue;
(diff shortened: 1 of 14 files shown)


mythz:
Can you please also include tests for all the new attributes/features supported, I'm going to rewrite the SQL generation so it shares as much impl as possible and want to be able to verify that it works after refactoring.

madaleno:
I will.
Currently struggling with VS as it doesn't want to compile ServiceStack.OrmLite.Tests.
Getting CS0006 errors about metadata file missing for every provider.
And since I am not a C#/VS developer (yet) it takes a little more time for me to solve some basic stuff sometimes.

I tested this by creating a new test project just for this purpose and it worked pretty well in SqlServer.
In FireBird it does not return values probably due to the lack of support in FireBird .net client.
Can only test in FireBird and SqlServer.
Will also investigate if Sqlite supports the returning clause.

mythz:
If you've already created a test project for this can you upload it to GitHub and I'll import it on my side, thx.

madaleno:
Hi @mythz .
Please check a sample test project for this feature here: 
[https://github.com/madaleno/ReturnAttributeTests](url)

mythz:
I might be missing something but where is the Sequence used in SQL Server? Also I notice in your [SqlServerTests.cs](https://github.com/madaleno/ReturnAttributeTests/blob/master/src/SqlServerTests.cs) you're using `db.CreateTable<User>(false)`, why aren't you re-creating the tables per test?

madaleno:
I was using a table already prepared and with data.
My table was created in a way that the Id value is calculated by the database on record insert to avoid gaps.
I don't like to get the sequence value inside my code and then pass that value to the insert command.
If something fails between the two you end up with gaps in the Id values.
In FireBird I always use a trigger to get the Id value from the corresponding sequence.
In SqlServer I use the default value: 
`[Id] bigint NOT NULL DEFAULT NEXT VALUE FOR [some_sequence_Id]`

Also, just notice that the SqlServer provider doesn't have support for sequences.
I will add it today and improve the test project.

mythz:
Tests needs to be reproducible, i.e. they can't rely on an external existing RDBMS tables. Also OrmLite itself as a code-first ORM should be able to re-create the table schemas using just the POCOs.

If the primary reason for this feature is to support sequences, the ideal API for this would be to allow users to decorate tables with:

```csharp
public class Table
{
    [Sequence("NameOfSequence")]
    public int Id { get; set; }
}
```

Where the Sequence will be created if it doesn't already exist and the `[Return] Id` metadata automatically inferred.

madaleno:
Yes. I am doing it in the test project.
But first I need to implement support for sequences in the SqlServer provider.

mythz:
The `[Default]` value of Sequences should be implied, why would you have a `[Sequence]` attribute on a property and did not want to use it? What are you expecting the behavior to be if no `[Default]` attribute existed?

If customizations are needed on Sequences they should be added on the `[Sequence]` attribute instead.

madaleno:
If you want to create a well defined table that can be safely used by others, we would define the default to get the value from the sequence.
In this case OrmLite would create the table so the Id field with a default to get the value from a sequence.
With this you would not need to use the Sequence attribute and the table would be prepared should any other app or process inserts data on it.

Implementing Sequences in the SqlServer, the code already produces sql to insert that includes the "NEXT VALUE FOR [NameOfSequence]" in the value part avoiding the need to do two db calls.

The example code above should produce this sql statement:
```sql
INSERT INTO USERS (Id, Name) VALUES (NEXT VALUE FOR "Gen_Users_Id", @name); 
```
Still need to change the create table code in the SqlServer provider to not create the Id column as IDENTITY(1,1) when it has the sequence attribute.

madaleno:
Ok, it's done now.
SqlServer dialect provider now supports Sequences.
@mythz can you take a look to my last commit?
The test project now includes the sequence attribute.

mythz:
Looks ok from here, will be able to tell once I had a closer look.

BTW do you really need `[ReturnOnUpdate]`? As this feature adds [3 Return Attributes](https://github.com/ServiceStack/ServiceStack/blob/master/src/ServiceStack.Interfaces/DataAnnotations/ReturnAttribute.cs) and I don't see `[ReturnOnUpdate]` implemented or tests for it in which case we only need the `[Return]` attribute which is only used in INSERTS. 

I also need to provide docs/examples on each feature and can't tell the difference between `[Return]` and `[ReturnOnInsert]`, also I can't think of a valid use case for `[ReturnOnUpdate]` which I don't expect to be ever used so I think we should just drop the attributes altogether.

madaleno:
Well Return was to be for both inserts and updates.
And the other two for each case. That was the plan but I thing that implementing this feature in update statements won’t be easy if possible at all because updates would return multiple rows.
So for now we should keep only one return attribute: Return or ReturnOnInsert.
When updating one row only we can easily do a select after, to get the complete updated object. So having this feature in updates might not be that important.

I will change the code to use only one attribute.

madaleno:
Ok. Did the ReturnAttribute cleanup.
Also fixed some bugs and added one more test and included the test class in the SqlServerTests project. 
Can you please take a look now.
Thanks.

mythz:
thx for the PR. 

I made some minor adjustments in https://github.com/ServiceStack/ServiceStack.OrmLite/commit/16ac261ecfb2ee5414c2c0fffbaed0c82a970e58

Ended up reverting back to `[ReturnOnInsert]` as I agree it's more descriptive as for its purpose. Note the tests didn't work with `[AutoIncrement]` (Requires IDENTITY_INSERT) but this shouldn't be needed with Sequences as a Sequence on the PrimaryKey should mean that the Sequence gets used.

Also the `[PrimaryKey]` is implied for `Id` properties so to use a sequence that returns on Insert can be reduced to:

```csharp
public class User
{
    [Sequence("Gen_UserSequence_Id"), ReturnOnInsert]
    public int Id { get; set; }

    public string Name { get; set; }
    public string UserName { get; set; }
    public string Email { get; set; }
}
```

If needed this is now available from v5.0.3 that's now [available on MyGet](http://docs.servicestack.net/myget).

PR #612, opened 2018-06-12
 Using GetFieldValue in InsertOnly just like in Insert

OlegNadymov:
This is critical for inserting boolean values into the Sqlite database using the InsertOnly method.


mythz:
Great thx, Oleg!

OlegNadymov:
@mythz thank you!

PR #613, opened 2018-06-16
T4 mapping improvement for SqlServer 

rsafier:
Add proper encoding of TIME(n) SqlServer fields to TimeSpan in T4 templates (still requires SqlServerTimeConverter to operate correctly)
Apply automatic substitution of valid SqlServer tables characters to valid .NET type name.


mythz:
Great stuff, thx Richard 👍

PR #621, opened 2018-11-04
SqlServerJsonStringConverter performance issue

drakelord:
Hello, recently, after switching to SqlServer 2016, I noticed performance issue while selecting big data sets, approximately 10-15 times slower.
It turned out to be SqlServerJsonStringConverter's fault which is added in SqlServer2016OrmLiteDialectProvider. It [uses](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/2147e466032cd23893b0d43137b06cfde1f51ed0/src/ServiceStack.OrmLite.SqlServer/Converters/SqlServerJsonStringConverters.cs#L15) [HasAttribute](https://github.com/ServiceStack/ServiceStack.Text/blob/1a6d461a72e2149a2c2f72484d7f274327702d08/src/ServiceStack.Text/PlatformExtensions.cs#L123) from ServiceStack.Text.PlatformExtensions and reselecting type's all attributes on each converter call. I think it's better to add cached version of HasAttribute to ServiceStack.Text.PlatformExtensions, so this PR is more to point out problem then real PR. I can make PR to ServiceStack.Text if my version of cached HasAttribute is acceptable.

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer/Converters/SqlServerJsonStringConverters.cs
+++ b/src/ServiceStack.OrmLite.SqlServer/Converters/SqlServerJsonStringConverters.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Data;
+using System.Linq;
 using System.Data.SqlClient;
 using ServiceStack.DataAnnotations;
 using ServiceStack.OrmLite.Converters;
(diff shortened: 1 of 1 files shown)


mythz:
I've decided to create an optimal cached version of `HasAttributeCached` in ServiceStack.Text so it's available to all in this commit: https://github.com/ServiceStack/ServiceStack.Text/commit/b0884b8d4eef3f20af482c69646f460e2678a43b

Then replaced `SqlServerJsonStringConverter` to use the optimal `HasAttributeCached` in this commit: https://github.com/ServiceStack/ServiceStack.OrmLite/commit/359bc1271dd7b37b8bb640c2fcb7149bb84b01e6

This change is available from v5.4.1 that's now [available on MyGet](http://docs.servicestack.net/myget).

drakelord:
> What’s the reason for using TypeHandle instead of the Type as the key?
No particular reason, both are valid as far as I know. I just personally prefer struct as a key.

mythz:
Apologies, updated in [this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/0b87ada03403b8f8e3da1b672ea6554b4f3f506e).

Now available in v5.4.1 on MyGet.

PR #622, opened 2019-01-14
[WIP - Do not merge] Solving some inconsistencies regarding tables naming convention

thibault-reigner:
It seems the code is mixing the use of ModelDef.ModelName and the value defined by INamingStrategy/IOrmLiteDialectProvider.GetTableName to get the name of a table.
Most of the time, the convention defined by the naming strategy is ignored under the assumption that INamingStrategy.GetTableName(ModelDef) is always equal to ModelDef.ModelName.

As a matter of example, let say I define a table through the following class
```
class Relation<A,B> 
{
// ....
}
```
with a naming strategy ensuring that this kind of class will get the name "AtoB". 
Depending on whether NamingStrategy.GetTable(modelDef) or ModelDef.ModelName is called in the code, I may end up with either "AtoB" or "Relation`2".

The changes I have performed in this PR are just a sketch because the whole structure of the API (mixing calls with either a ModelDefinition or a string as parameter) does not make it easy to normalize.

What should be the best approach here to fix this issue ?

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.Join.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.Join.cs
@@ -183,7 +183,7 @@ namespace ServiceStack.OrmLite
             if (refField == null) 
             {
                 if(!isCrossJoin)
-                    throw new ArgumentException($"Could not infer relationship between {sourceDef.ModelName} and {targetDef.ModelName}");
+                    throw new ArgumentException($"Could not infer relationship between {DialectProvider.GetTableName(sourceDef)} and {DialectProvider.GetTableName(targetDef.ModelName)}");
 
                 return string.Empty;
             }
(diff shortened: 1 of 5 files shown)


mythz:
If you want a name that's not the class name use an `[Alias]` to specify the name you want to use.

Note: OrmLite has never supported using generic types for tables, you can use generic types for base classes but it was never designed/tested to work to use generic types for POCO table classes.

If you want to make any changes, you'll also need to include tests to validate why the change is needed.

thibault-reigner:
Noted, I will proceed to a dedicated PR with new unit tests. However I would like to highlight the fact that the use of generic types is actually irrelevant to the issue. I could simply enforce a convention like prefixing the name of the tables by a specific string, the issue would be the same : 
The API suggest the possibility to set the naming convention through the injection of a INamingStrategy  in OrmLiteDialectProviderBase but internally this INamingStrategy is arbitrarily ignored.

PR #624, opened 2019-03-11
Dockerized db server instances

wwwlicious:
@mythz So I now have the initial dockerized instances of Firebird, MySql (mariaDb), Postgres, Oracle XE and MS SQL running tests. The list is in the [readme](https://github.com/wwwlicious/ServiceStack.OrmLite/blob/docker-db/src/Docker/readme.md)

There are plenty of tests still failing which I think have always been failing for me when installing full local db servers so I need to work through these. Some haven't been maintained such as Oracle and I don't think they have been run for quite a while.

Let me know what you think and if you recommend any changes.


mythz:
@wwwlicious fantastic stuff, thx!

It looks like all failed tests are mostly due to the SQL Server connection string changing, i'll update it so it looks for the connection string in an Environment variable first if it exists and restore the existing connection in our CI.

Long term goal would be change our CI to test against these docker instances, so many thx for this!

wwwlicious:
@mythz no worries. Going to look next at how/if I can re-run the tests for multiple db versions and tidy up some of the multiple config strings scattered around. 

Also I want to see if there is scope to use nunit categories to skip tests where appropriate (I tend to think in xunit terms) and automatically spin up and teardown db instances.

Also been a while since I actively used TC so not sure what the deal is with caching docker images, would welcome any input you have on getting a good balance between local and CI testing.

mythz:
For the time being it's just being able to [substitute the connection strings used](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/180ee75613b5c73dd4f88e3e46e960bf9509e352) which did the trick as now OrmLite CI is passing again.

I'll have to revisit using the Docker images after the next release.

PR #626, opened 2019-04-01
Test refactor and fixes

wwwlicious:
@mythz  again I open this more for discussion than merging.

Would recommend you checkout the branch, spin up docker and run the main test proj which should all pass. (sqlite by default)

Then set the dialect flags as the test runner environment variable DefaultDialects = Supported and rerun tests. Should discover north of 13k tests per target framework.

The outstanding failures indicate either non-db agnostic tests and there are some issues with schema handling in the likes of postgres and mysql/mariadb where the schema is appended to the table name rather than actually using schemas in these db's and this behaviour is not consistent across all api's for the dialectproviders.

Anyway, have a look and let me know your thoughts on the general approach.

The test [readme](https://github.com/wwwlicious/ServiceStack.OrmLite/tree/oracle-managed/tests) covers the main bits


mythz:
Schemas in MySql aren't the same as databases, they're effectively a synonymous with database whereas they're typically a "namespace" within a Database for other RDBMS's. 

I believe it's the only external environment condition that the tests depend on, that there's an existing Schema called "Schema" for RDBMS's that support Schemas. For other RDBMS's we fake it by including the schema name as part of the table name.

For switching Dialects I'd prefer we have a `ORMLITE_DIALECT` Environment which overrides `Config.DefaultDialect` if it's defined as it makes it easy to create different CI projects with different Environment variables.

mythz:
I've got the [docker-compose RDBMS's instances up](https://github.com/ServiceStack/ServiceStack.OrmLite/tree/master/src/Docker) easy enough (nice job BTW) but couldn't get any of the filters to work.

I'm running Docker on Linux so I've had to disable `net46`, trying to run `--filter TestCategory=AnySqlServer` results in:

```
root@Ubuntu-1604-xenial-64-minimal ~/src/ServiceStack.OrmLite/tests/ServiceStack.OrmLite.Tests # dotnet test --filter TestCategory=AnySqlServer
Build started, please wait...
Build completed.

Test run for /root/src/ServiceStack.OrmLite/tests/ServiceStack.OrmLite.Tests/bin/Debug/netcoreapp2.1/ServiceStack.OrmLite.Tests.dll(.NETCoreApp,Version=v2.1)
Microsoft (R) Test Execution Command Line Tool Version 15.9.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
No test matches the given testcase filter `TestCategory=AnySqlServer` in /root/src/ServiceStack.OrmLite/tests/ServiceStack.OrmLite.Tests/bin/Debug/netcoreapp2.1/ServiceStack.OrmLite.Tests.dll
```

Basically same with everything else I've tried, e.g:

```
No test matches the given testcase filter `TestCategory=PostgreSql9` in /root/src/ServiceStack.OrmLite/tests/ServiceStack.OrmLite.Tests/bin/Debug/netcoreapp2.1/ServiceStack.OrmLite.Tests.dl
```

But was able to run it without the filter:

```
$ dotnet test
....
Total tests: 993. Passed: 951. Failed: 1. Skipped: 35.
Test Run Failed.
Test execution time: 14.4456 Seconds
```

I'm assuming that's just running SQLite by default right? Anyway it would be nice to be able to select the Dialect based on a custom `ORMLITE_DIALECT` Environment variable as that would be easiest to configure having multiple CI project clones which are just called with a different environment variable.

wwwlicious:
> Schemas in MySql aren't the same as databases which are synonymous with database whereas they're typically a "namespace" within a Database for other RDBMS's.

Yes, I know, my point was that the prefixing behaviour isn't consistent across all the api's which accounts for a few test failures

Another area where there are a few test failures is casing in Postgres. I wasn't sure of the best way to tackle this as I didn't want to change behaviour of existing impl but it might be an idea to have a flag for naming strategy that uses the backtick in all identifiers to preserve casing. It becomes clearer once you see how some of the tests fail against postgres.

> For switching Dialects I'd prefer we have a ORMLITE_DIALECT Environment which overrides Config.DefaultDialect if it's defined as it makes it easy to create different CI projects with different Environment variables.

Done. The variable that controls what dialects are run is `TestConfig.DefaultDialects`
Is does run Sqlite by default, the value is set to sqlite but overridden by an environment variable (now renamed or ORMLITE_DIALECT. This is then injected for each distinct database Dialect flag value into the ctor of the tests.

I think the filters can only reduce/opt out of dialects tests rather than include them as it isn't straightforward to detect filters in the test assembly at runtime, so the env var or modifying `TestConfig.DefaultDialects` is the primary driver for establishing what tests are discovered to run as you mentioned anyway. (Will update readme to be clearer on this)

I also noticed last night that I should make the test setup that depends on the docker instances optional so that the sqlite tests can run without docker dbs being spun up (the reason the build on this PR failed).

mythz:
> I also noticed last night that I should make the test setup that depends on the docker instances optional 

Yeah it should be able to run locally (SQLite default) and have an easy way to change the Dialect + connection string that's used i.e. so we can use any RDBMS instance (like the ENV connection string allows).

I think that's all we need before we can merge right?

wwwlicious:
@mythz I'll look to make those changes. 

As for merging, there are still around 220 failing tests of the 13k+ across the officially supported providers/db versions which are a mixture of the tests themselves being provider specific and provider bugs/inconsistent behaviours across the api's. I think the former I should fix first.

Also the branch is named 'oracle provider' as that is where I originally started porting the devart provider over to the oracle managed one and I haven't completed that yet. It has proved challenging to even get different versions of XE into the docker-compose due to various DMCA issues/oracle account logins etc  

Its ultimately your call on merging if you are comfortable that this breaks some things for the providers not officially supported and that the testing reveals a number of bugs to fix. (I fixed a few already tho 😀)

There is also a large tidy up job to be done on amount of test duplication where the same api's are tested 
in different files for each provider, something the new base class eliminates the need for. In fact most of the test projects for the other providers are either redundant or could be reduced to only db specific feature tests which themselves could be merged into the main test project as there is sufficient granular control over which tests run for which provider at project, class and test levels.

As I said initially, my intent on this was really just to get your thoughts on the direction so far.

mythz:
I ran the primary [ServiceStack.OrmLite.Tests](https://github.com/ServiceStack/ServiceStack.OrmLite/tree/master/tests/ServiceStack.OrmLite.Tests) project and there was only 1 failing test for SQLite which I should be able to resolve if I can repro it here. They're the only test suite I that we need to run in CI initially.

I don't mind if there are failing tests in the other RDBMS-specific test projects, they're primarily adhoc integration tests to test behavior in specific providers, they weren't part of the official test suite and likely have a lot more integration/environment dependencies and likely to be using class names that conflict with different types with the same name the main test suite.

wwwlicious:
@mythz ok, I already had the env vars to override the connectionstrings in [TestConfig](https://github.com/wwwlicious/ServiceStack.OrmLite/blob/oracle-managed/tests/ServiceStack.OrmLite.Tests/TestConfig.cs#L14)

I've updated the base class so that the default sqllite tests run when docker isn't being used.

mythz:
Great, going to take the CI's green tick as a green light :)

wwwlicious:
Haha 🙈 you're a brave soul!

mythz:
@wwwlicious Hey Scott just want to provide an update.

I'm now running the CI against the latest version of the different supported RDBMS's. I've checked in the [/servicestack-ci/docker-compose.yml](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/src/Docker/servicestack-ci/docker-compose.yml) that ServiceStack CI Server is using for all projects.

There's a test project for each of the supported RDBMS's (SQLite,SQL Server, PostgreSQL, MySQL), 1 for .NET Framework/Windows and 1 for .NET Core/Linux. 

I've also added .NET FX/.NET Core test projects for Oracle as well but as there's so many tests failing I've switched the Windows project to Manual since it will always fail until someone wants to fix the Oracle provider. There's only .NET Core for Firebird for same reasons.

BTW thanks for all your efforts + restructuring, didn't know about the NUnit features to allow running normal tests multiple times which makes it easy to be able to run against multiple RDBMS's and versions in one test run. I'm not using the feature atm as each integration test runs 1 RDBMS against the latest version available.

I've made some refactoring where I've changed [Dialect providers](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/tests/ServiceStack.OrmLite.Tests.Setup/TestConfig.cs) to map 1:1 with different available dialect providers. Each RDBMS can test multiple versions which can be overridden with an environment variable (e.g. `MYSQL_VERSION`) of comma-delimited version numbers. 

I've also moved the RDBMS setup scripts into a separate [ServiceStack.OrmLite.Tests.Setup](https://github.com/ServiceStack/ServiceStack.OrmLite/tree/master/tests/ServiceStack.OrmLite.Tests.Setup) test project which is run before each CI test project. In my tests 1s wasn't enough time as it was often timing out before the tests were run so I've changed it to 10s.

But otherwise everything else should be the same, all supported RDBMS run locally for me, but there's 1 test on MySQL still failing which I can't repro locally yet. Hopefully all my changes are easy to follow.

wwwlicious:
@mythz yes, I've been seeing the commits, think I follow it all and the changes look great. 👍

For running the tests locally, it was convenient to be able to set the env var, spin up the db's and run all tests against all providers but it makes sense to split this apart for CI. 

There are a few errors I found when running dialects against older RDBMS versions (MySql 5.5 and datetime conversion for example), could be worth some testing against these. 

I'll pick up the oracle provider shortly

PR #628, opened 2019-05-23
FireBird 4 support

madaleno:
Added support for FireBird 4 new Identity columns.
Fixed create table and insert row sql then there are fields with AutoId attributes.
Changed behavior of AutoIncrement fields to use Identity instead of Sequences when using the new FireBird 4 dialect.
Added some tests.


mythz:
Cheers 👍

PR #629, opened 2019-05-23
More on FireBird4.

madaleno:
1. Updated reserved words in FireBird4 dialect provider.
2. Changed the autoincrement definition from "by default" to "always" since there is no easy way to generate the insert sql based on object values. To use "By Default" the generated insert sql should not contain the Id field name if the Id value is null.
3. Updated FirebirdClient package to 6.6.0.
4. Changed obsolete method SelectFmt in Schema.cs.


mythz:
Thanks 👍

PR #630, opened 2019-06-02
 Fix for OrmLite.Save().

madaleno:
Fixed OrmLite.Save() that was always returning false when the object already exists in the database. 
In those cases Save() calls Update() that returns the number of records affected but Save() was returning false ignoring the result of the Update() call.


mythz:
Sweet, thx. 👍 

Since you're doing so much with Firebird4 Support I think we should rename the existing `FirebirdOrmLiteDialectProvider` provider to `Firebird3OrmLiteDialectProvider` and this provider to `FirebirdOrmLiteDialectProvider` so it's the default, what do you think?

madaleno:
Hi.

I tried to follow what was done in the SqlServer provider.
I don't think it is a good idea, at least now, because it would create a breaking change.
And also because the existing provider is not just for FireBird 3 and the name could be confusing.

Thanks.

PR #637, opened 2019-10-31
HasAttributeCached perfomance issue

OlegNadymov:
I got performance problem after update OrmLite to version 5.7.0. I found the method that now works longer: ``OrmLiteConfigExtensions.GetModelDefinition``. I suppose the problem is in ``HasAttributeCached`` method that used in ``GetModelDefinition``. Also I changed ``HasAttribute`` to ``HasAttributeCached`` in my project and it also got slower.
I created two tests where you can see 5 times performance drop.
Could you fix this performance problem? I think it is very important not only for me.

### DIFF
--- a/tests/ServiceStack.OrmLite.Tests/OrmLiteConfigExtensionsTests.cs
+++ b/tests/ServiceStack.OrmLite.Tests/OrmLiteConfigExtensionsTests.cs
@@ -1,0 +1,140 @@
+﻿using NUnit.Framework;
+using ServiceStack.Common.Tests.Models;
+using ServiceStack.DataAnnotations;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Serialization;
+using static ServiceStack.OrmLite.Tests.Issues.LoadSelectIssue;
+
+namespace ServiceStack.OrmLite.Tests
+{
+    [TestFixture]
+    public class OrmLiteConfigExtensionsTests
+    {
+        private readonly List<Type> _attributes = new List<Type>{
+            typeof(PreCreateTableAttribute),
+            typeof(PostCreateTableAttribute),
+            typeof(PreDropTableAttribute),
+            typeof(PostDropTableAttribute),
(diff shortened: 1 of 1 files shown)


OlegNadymov:
@mythz  I'm wondering why CI has failed checks.

mythz:
I'm not seeing the **2600-2800 ms** perf issue indicated in your comment, running .NET Core tests on the command-line 

    $ dotnet test --filter OrmLiteConfigExtensionsTests --verbosity normal

reports:

```csharp
NUnit Adapter 3.15.1.0: Test execution complete

  V HasAttribute [24ms]

  V HasAttributeCached [16ms]
```

Running this in Rider's NUnit test runner in .NET v4.6.1 reports 16ms for `HasAttributeCached`:

image

Although since your test is not testing perf for retesting the same entry I'd expect the non-cached version to be a bit slower, so IMO the perf difference is normal run variance, probably faster because `HasAttributeCached` is run second (and chance to warm up).

Maybe you're using an older version of ServiceStack.Text as I've recently [replaced HasAttributeCached 29 days ago](https://github.com/ServiceStack/ServiceStack.Text/commit/628a4a181ec09c0f68daa91332d7feccf776fc59) to use a ConcurrentDictionary due to a perf regression with re-building the non-threadsafe dictionary after each new entry.

If that's the case [clear your NuGet packages cache](https://docs.servicestack.net/myget#redownloading-myget-packages) to download the latest version of ServiceStack.Text.

OlegNadymov:
@mythz Was the new version of ServiceStack.Text released to NuGet? I cleared cache of NuGet packages but anyway ServiceStack.Text has remained old version with the non-threadsafe dictionary.
Or do you mean MyGet feed (https://www.myget.org/F/servicestack) ? If so, when the new version will be released to NuGet?

mythz:
You'd need to download the v5.7.1 packages on MyGet, there's no NuGet release planned for a while.

OlegNadymov:
I got you. Thanks!

OlegNadymov:
The problem was gone after I updated to v.5.7.1. Thanks!

PR #639, opened 2019-11-19
Fix RowCount not working with SelectFilters.

timzap:
`countExpr` is a clone of `expression` so it did not gain any new parameters added by `.ToSelectStatement()`, which was causing an exception when being executed.


mythz:
Great, thx for the PR + tests!

mythz:
FYI if needed this change is available from the latest v5.7.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #641, opened 2020-02-05
Include appropriate files in ServiceStack.ormLite.SqlServer.Data.Core

danielmarbach:
See twitter discussion https://twitter.com/danielmarbach/status/1224952168563838976

When referencing ServiceStack.OrmLite.SqlServer.Data to get access to Microsoft.Data.SqlClient the full framework target does not properly resolve the dependencies, It only ships netstandard2.0 but its dependencies have net45;netstandard2.0. This causes dependency issues. When you reference that package on net462 the package brings in netstandard but the dependencies go for net45 assemblies which then causes build errors like the following

image
image

It was then recommended to use ServiceStack.OrmLite.SqlServer.Data.Core which is netstandard only. Unfortunately if I pull this in I'm not able to set

```
OrmLiteConfig.DialectProvider = SqlServerDialect.Provider;
```

With a custom build containing the PR I can now properly use the SqlServerDialect

### DIFF
--- a/src/ServiceStack.OrmLite.SqlServer.Data/ServiceStack.OrmLite.SqlServer.Data.Core.csproj
+++ b/src/ServiceStack.OrmLite.SqlServer.Data/ServiceStack.OrmLite.SqlServer.Data.Core.csproj
@@ -22,4 +22,70 @@
     <PackageReference Include="Microsoft.Data.SqlClient" Version="1.1.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\ServiceStack.OrmLite.SqlServer\Converters\SqlServerBoolConverter.cs">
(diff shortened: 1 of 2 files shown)


danielmarbach:
Thanks man! There is no hurry but would it be possible to release a patch at some time?

mythz:
For some reason this .csproj update broke MSBuild where I ended up having to extract it out from build-core.proj into a separate [build-core-data.proj](https://github.com/ServiceStack/ServiceStack.OrmLite/blob/master/build/build-core-data.proj) MSBuild project and use `dotnet` to pack it as the MSBuild task kept failing.

Anyways this package is now [available from v5.8.1 on MyGet](https://docs.servicestack.net/myget), it wont be available on NuGet till next major release which wont be for a while.

PR #642, opened 2020-02-25
Update to find latest compatible MSBuild for creating building projects

Siliconrob:
Building locally there is a hard dependency to Visual Studio 2017.  This change is to find the latest MSBuild and then build from that particular exe


mythz:
Cheers 👍

PR #644, opened 2020-03-13
BUGFIX: Add schema prefix to inner join when present

WardenUnleashed:
When I was attempting to select from an InnerJoin db query of two table classes with `RowVersion` properties and a `Schema` attributes, I started receiving an exception.

I'm not sure if this error was introduced with the reworkings of RowVersion that have been done in the last week or perhaps the use-case has not been encountered before but it would appear that the schema for the table was not being appended when selecting the RowVersion column.

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.Join.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.Join.cs
@@ -356,7 +356,7 @@ namespace ServiceStack.OrmLite
                                 }
                                 else
                                 {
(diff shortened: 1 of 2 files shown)


mythz:
Cheers 👍 thx for the fix + test!

WardenUnleashed:
> Cheers 👍 thx for the fix + test!

No problem! Any ETA for when this fix will go out to Nuget?

mythz:
It’s available from [v5.8.1+ on MyGet](https://docs.servicestack.net/myget) now, won’t be on NuGet till next major release which won’t be for a while.

PR #645, opened 2020-04-09
Add NativeTypes int[] alias and bool[] boolean[]

cthames:
Added int[] alias to match Postgres alias. Also added missing boolean array types.


mythz:
Cheers 👍

PR #646, opened 2020-04-09
Fix test execute stored procedure w/ array args

cthames:
Fixes the test and removes the test [Ignore] attribute.
Also adds additional execute option, with and without [Alias("")] attribute.


cthames:
Probably should have put a comment on the reason why it wasn't working.

"If CommandType.StoredProcedure is set... your NpgsqlParameter names must match your PostgreSQL function parameters."

Relevant info from: https://www.npgsql.org/doc/basic-usage.html
>Note that **if CommandType.StoredProcedure is set** and your parameter instances have names, Npgsql will generate parameters with named notation: SELECT my_func(p1 => 'some_value'). This means that your **NpgsqlParameter names must match your PostgreSQL function parameters**, or the function call will fail. If you omit the names on your NpgsqlParameters, positional notation will be used instead. See the PostgreSQL docs for more info. https://www.postgresql.org/docs/current/static/sql-syntax-calling-funcs.html

mythz:
Great, thx.

PR #647, opened 2020-05-14
Delete command filter

hugoj-goncalves:
Delete does not have a command filter like Update and Insert methods, would be cool to have also.
I'm not sure if those are the best places to call the filter

### DIFF
--- a/src/ServiceStack.OrmLite/OrmLiteWriteExpressionsApi.cs
+++ b/src/ServiceStack.OrmLite/OrmLiteWriteExpressionsApi.cs
@@ -234,9 +234,9 @@ namespace ServiceStack.OrmLite
         ///   db.Delete&lt;Person&gt;(p => p.Age == 27);
         ///   DELETE FROM "Person" WHERE ("Age" = 27)
         /// </summary>
-        public static int Delete<T>(this IDbConnection dbConn, Expression<Func<T, bool>> where)
+        public static int Delete<T>(this IDbConnection dbConn, Expression<Func<T, bool>> where, Action<IDbCommand> commandFilter = null)
         {
-            return dbConn.Exec(dbCmd => dbCmd.Delete(where));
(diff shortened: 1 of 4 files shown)


src/ServiceStack.OrmLite/OrmLiteWriteExpressionsApi.cs
```diff
@@ -246,7 +246,7 @@ public static int Delete<T>(this IDbConnection dbConn, Expression<Func<T, bool>>
         ///   db.Delete&lt;Person&gt;(q.Where(p => p.Age == 27));
         ///   DELETE FROM "Person" WHERE ("Age" = 27)
         /// </summary>
-        public static int Delete<T>(this IDbConnection dbConn, SqlExpression<T> where)
+        public static int Delete<T>(this IDbConnection dbConn, SqlExpression<T> where, Action<IDbCommand> commandFilter = null)
         {
             return dbConn.Exec(dbCmd => dbCmd.Delete(where));
```
madaleno, line 251: shouldn't this be: ... dbCmd.Delete(where, commandFilter)); ?

mythz: Yep it should be, thx for catching it! I'll resolve it on my side.

mythz:
FYI made minor tweaks and added equivalent support for Delete Async APIs in [this commit](https://github.com/ServiceStack/ServiceStack.OrmLite/commit/e818b16722b369742cc3986f07b83df3d730848a).

This needed these changes are now [available on MyGet](https://docs.servicestack.net/myget).

src/ServiceStack.OrmLite/OrmLiteWriteExpressionsApi.cs
hugoj-goncalves: Ops

PR #651, opened 2021-01-24
add ToString format paramater support for MySql DateTime

pedoc:
currently, ```ToString``` on ```DateTime``` will ignore any format parameters and always convert to ```cast({quotedColName} as char(1000))```,but most developers (at least in our opinion) want to convert according to the specified format and use it in ```Where``` or ```Select```.

this commit achieves this goal by passing the ```ToString``` parameter as it is to the ```DateFormat``` function of Mysql.

### DIFF
--- a/src/ServiceStack.OrmLite.MySql/MySqlExpression.cs
+++ b/src/ServiceStack.OrmLite.MySql/MySqlExpression.cs
@@ -1,9 +1,12 @@
 ﻿namespace ServiceStack.OrmLite.MySql
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Linq.Expressions;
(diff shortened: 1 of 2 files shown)


mythz:
Hi, thx for the PR! Can you also add a new test in [ServiceStack.OrmLite.MySql.Tests](https://github.com/ServiceStack/ServiceStack.OrmLite/tree/master/tests/ServiceStack.OrmLite.MySql.Tests) with an example of what this PR allows, thx.

pedoc:
@mythz hi,i added some tests, please review,thx

mythz:
Much Appreciated, thx!

This PR is now available from v5.10.5+ that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #652, opened 2021-02-24
To insert statement dbnull check

Layoric:
Fix for when table class has reference type (treated as TEXT or JSON column) and is provided null. Previously generated `{}` for null values with if inserted and then retrieved results in different instance values.

Added 2 tests.


mythz:
Looks good 👍

PR #653, opened 2021-05-26
Fix select filed alias conflict bug of main table

wuyou331:


mythz:
thx for the PR! 

This change is now available from v5.11.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #654, opened 2021-07-05
Fix in InsertOnly for param greater than default length

OlegNadymov:
``InsertOnly`` had different logic than the ``Insert`` method when handling parameter values. Now ``InsertOnly`` also uses ``dialectProvider.SetParameterValues``.


OlegNadymov:
@mythz  If I get it right the "Nuget restore" step is failed. Could you fix it?

mythz:
thx for the PR 👍 This is now available from the latest v5.11.1 that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #657, opened 2021-09-06
Improve the implementation of FloatConverter to fix the conversion failure problem in some cases.

pedoc:
In some cases, the default ```FloatConverter``` implementation will throw an ```InvalidCastException``` exception, sample code: 
```
 var floatConverter = new LegacyFloatConverter();
 floatConverter.ToQuotedString(typeof(decimal), 89.123456789);
```

This fix uses the function of the Convert class to replace the forced conversion to solve this problem.For details, please see changes and unit tests.

thanks

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,17 +1,20 @@
 {
     // See https://go.microsoft.com/fwlink/?LinkId=733558
     // for the documentation about the tasks.json format
-    "version": "0.1.0",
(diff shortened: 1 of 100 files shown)


mythz:
Appreciated, thx 👍

mythz:
FYI, if needed this change is now [available on MyGet](https://docs.servicestack.net/myget).

PR #661, opened 2021-11-22
Fix the 'EndExecuteNonQuery' asynchronous concurrent execution

vla:
UntypedApi Asynchronous API concurrency problem

### DIFF
--- a/src/ServiceStack.OrmLite/UntypedApi.cs
+++ b/src/ServiceStack.OrmLite/UntypedApi.cs
@@ -48,6 +48,11 @@ namespace ServiceStack.OrmLite
             return DbCmd != null ? filter(DbCmd) : Db.Exec(filter);
         }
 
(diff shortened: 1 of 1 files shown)


mythz:
Hi, thx for the PR 👍 Is there something missing in this commit, i.e. how will adding an API fix anything if it's not being used? Is it catching an existing method overload or something?

vla:
`UntypedApi.SaveAllAsync & SaveAsync` 
These two functions cause the concurrent execution of EndExecuteNonQuery to raise an error

vla:
```charp
public TReturn Exec<TReturn>(Func<IDbCommand, TReturn> filter)
        {
            return DbCmd != null ? filter(DbCmd) : Db.Exec(filter);
        }
```
The method async/await is invalid

mythz:
ok, so the it is catching an existing method overload.

FYI this change is now available on the latest v5.13.3+ that's now [available on MyGet](https://docs.servicestack.net/myget).

PR #662, opened 2021-11-23
Support custom order of fields when creating table.

pedoc:
Support for custom field order through ```CustomFieldOrderAttribute ```.

Limits:
One of the problems is that other ```Attribute``` definitions in OrmLite are placed under ```ServiceStack.Interfaces```, but due to the lack of control over this assembly, they are temporarily placed in ```ServiceStack.OrmLite```, but I recommend changing to ```ServiceStack.Interfaces``` for consistency. 

PS:
In addition, the grammar of C#10 has been rewritten to support vs2019.


mythz:
Awesome PR thx 👍 

There's a lot in here though, it's not just the `CustomFieldOrderAttribute`, since it's just used to configure the order IMO it should just use .NET's existing `[DataMember(Order=1)]`.

Can you please add each feature in a separate PR, so they can be evaluated/accepted individually? I'm having a hard time trying to correlate the purpose, code, & tests for each feature.

pedoc:
> Can you please add each feature in a separate PR

This PR actually only does one thing, which is to increase the allowable custom field order.


> There's a lot in here though, it's not just the `CustomFieldOrderAttribute`, since it's just used to configure the order IMO it should just use .NET's existing `[DataMember(Order=1)]`. If it did more we could add a `[ConfigField(Order=)]` attr, which we could extend with different supported features, but if it's just order it should used .NET's existing attribute.

I have different opinions on this point. ``DataMember`` APIs are more used for serialization and deserialization.I believe this is already a consensus . But you reminded me that I think this additional Attribute is indeed a bit redundant. Now that there is a ``CustomFieldAttribute``, do you think it is appropriate to put ``Order`` in this class?

mythz:
> This PR actually only does one thing, which is tosupport custom field order.

Then can you do a fresh pull and a new PR so that it only contains these commits (and not the irrelevant ones) please.

Yeah `CustomFieldOrderAttribute` is too verbose for such a small feature. I've just added a new [ConfigFieldAttribute](https://github.com/ServiceStack/ServiceStack/commit/6a8cf8e06d65660b94d0f4d086534ea89233b3cd) that we can start using to customize the configuration of each field.

mythz:
FYI if needed, the `ConfigFieldAttribute` is available from v5.13.3+ that's now [available on MyGet](https://docs.servicestack.net/myget).

pedoc:
> Can you please add each feature in a separate PR, so they can be evaluated/accepted individually? I'm having a hard time trying to correlate the purpose, code, & tests for each feature.

hi, fixed from #663

pedoc:
for [ConfigFieldAttribute](https://github.com/ServiceStack/ServiceStack/commit/6a8cf8e06d65660b94d0f4d086534ea89233b3cd),
Can we directly add the ``Order`` field to ``ServiceStack.DataAnnotations.CustomFieldAttribute`` to reduce the number of new entities?

mythz:
I didn't want to break compatibility with the original `[CustomField(sql)]` Sql constructor so I've left it alone and [just added optional Order](https://github.com/ServiceStack/ServiceStack/commit/00ff67f5fa3abe9aafb5b7a2c9a1c673862a262c) with a default constructor so you can now use `[CustomField(Order = 1)]`.

pedoc:
Ok,When the #663 is completed, I will continue my work here, thank you for your cooperation. :)

mythz:
np, thx for your continued efforts on this. FYI new `[CustomField]` is now available from the latest v5.13.3+ that's now available on MyGet.

PR #665, opened 2021-11-29
Support custom order of columns when creating table

pedoc:
Equivalent to #662


pedoc:
sorry,i am late :)

mythz:
Much nicer, thx 👍

PR #667, opened 2022-02-18
Improve enum conversion when indirect member access, such as Sql.TableAlias(), etc.

pedoc:
Consider the following code snippet:
```csharp
using var db = OpenDbConnection();
var q = db.From<EntityMonitor>();
var stateType = StateType.完好;
q.Or<ChnStateResult>((x) => Sql.TableAlias(x.State, nameof(LinkResult)) == StateType.离线);
q.Or<ChnStateResult>((x) => x.State == stateType);

var sql = q.ToMergedParamsSelectStatement();
```
Since ``StateType.离线`` will be compiled as a constant, the SQL compiles to ：
```sql
WHERE `LinkResult`.State=0 OR `LinkResult`.State='完好'
```
but expected should be 
```sql
WHERE `LinkResult`.State='离线' OR `LinkResult`.State='完好'
```

We expect indirect member access functions like Sql.TableAlias to have the same effect, which is exactly what this PR does.（Full example is in unit test ）

### DIFF
--- a/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
+++ b/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs
@@ -36,7 +36,7 @@ namespace ServiceStack.OrmLite
         public string TableAlias { get; set; }
         public IOrmLiteDialectProvider DialectProvider { get; set; }
         public List<IDbDataParameter> Params { get; set; }
-        public Func<string,string> SqlFilter { get; set; }
+        public Func<string, string> SqlFilter { get; set; }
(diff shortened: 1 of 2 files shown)


mythz:
Excellent, much appreciated 👍

pedoc:
> Actually this repo is now longer maintained and has [moved to a mono repo](https://docs.servicestack.net/mono-repo) at:
> 
> https://github.com/ServiceStack/ServiceStack/tree/main/ServiceStack.OrmLite
> 
> Can you also submit this and any future PR's there please.

sorry my late :),
here now:https://github.com/ServiceStack/ServiceStack/pull/1271

PR #669, opened 2022-12-23
Npgsql is not compatible with version 7 and can result in runtime error

ramonsmits:
Resolves the following exception by limiting version of Npgsql:

> System.TypeLoadException: Could not load type 'NpgsqlTypes.NpgsqlDate' from assembly 'Npgsql,

This range should be included in the package reference.


mythz:
It's not expected for major versions to be compatible, which typically indicates major breaking changes. 
Not going to start adding major version constraints to any packages now, but will be looking at the effort required to support v7's breaking changes soon.

